PR #1, opened 2024-11-15
Add `.gitignore`

MDLC01:
The content is partially copied from [typst/typst](https://github.com/typst/typst).


laurmaedje:
Thanks! I forgot about that because my Rust target dir is system-wide.

PR #2, opened 2024-11-15
Add gender symbols (Proposal 1)

MDLC01:
## Description

This PR implements Proposal 1 (iteration ~2~ 3) from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). It adds gender symbols under a new `sym.gender` sub-module, ~and removes the old `emoji.transgender` symbol~.

~`emoji.transgender` is no longer valid.~ No breaking change as of iteration 3 (07bfd3134a695a990e9fd197be909fd7db28de0e).

### DIFF
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -1347,6 +1347,7 @@ train 🚆
   .suspend 🚟
   .tram 🚊
   .tram.car 🚋
+transgender ⚧
 tray
   .inbox 📥
   .mail 📨
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -1347,7 +1347,6 @@ train 🚆
   .suspend 🚟
   .tram 🚊
   .tram.car 🚋
-transgender ⚧
 tray
   .inbox 📥
   .mail 📨
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1026,3 +1026,18 @@ Im ℑ
 dotless
   .i ı
   .j ȷ
+
+gender {
+  female ♀
+    .double ⚢
+    .male ⚤
+  intersex ⚥
+  male ♂
+    .double ⚣
+    .female ⚤
+    .stroke ⚦
+    .stroke.t ⚨
+    .stroke.r ⚩
+  neuter ⚲
+  trans ⚧
+}


emilyyyylime:
[meta] Would a 'breaking change' label be useful?

emilyyyylime:
Also CC @laurmaedje; could we devise some united protocol with [typst/typst](https://github.com/typst/typst) for deprecating and eventually removing symbols?

laurmaedje:
> Also CC @laurmaedje; could we devise some united protocol with [typst/typst](https://github.com/typst/typst) for deprecating and eventually removing symbols?

Yeah, some sort of deprecation mechanism would be nice. But Typst doesn't even have that for any kind of constant yet, so we'd need to start there, maybe via some mechanism for deprecated entries of a `Scope`.

laurmaedje:
Not sure how to deal with those hybrid thingies, which are sometimes emojis and some not. Doesn't removing it from `emoji` reduce our emoji coverage? Perhaps we should rather opt for suporting variation selectors in codex (by exposing strings rather than chars probably?) and force one to the emoji and one to the text presentation.

Slightly related are modifiers for skin tone etc. which should probably not be added by hand.

Side note: If we want to support multiple codepoints that could also have an effect on #9.

MDLC01:
I agree it would probably make more sense in the long term to use clusters (probably exposed as strings) instead of codepoints. I just did not think it was an issue in this case.

If we do not decide to change that now, I see two options regarding this PR:
- Close it for now;
- Keep `emoji.transgender` for now, and add the appropriate variation selectors later.

laurmaedje:
I think I'd be okay with the latter, I just don't want to remove it if we're gonna add it back anyway.

MDLC01:
This PR is no longer breaking

mkorje (approved):

T0mstone:
For future reference, this is related to #25, since U+2640, U+2642 and U+26A7 seem to have emoji variation selectors. (Not sure about the others, I didn't consult an authoritative source and instead just tried adding VS16).

laurmaedje:
I should note that the current docs system will very likely choke on the fact that `gender` is a module.

MDLC01:
Ideally, submodules of `sym` and `emoji` should be listed in the General Symbols and Emoji Symbols pages. And their symbols should searchable like any other symbols. Not sure how easily to communicate that the "base" is a module rather than a symbol, though. Maybe submodules should have a separate list within the General Symbols and Emoji Symbol pages, but then they would not be searchable as easily.

Enivex (approved):

PR #3, opened 2024-11-15
Add die face symbols (Proposal 2)

MDLC01:
This PR implements Proposal 2 (iteration 1) from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). It adds die face symbols as `die.{one,two,three,four,five,six}`, with `die.six` being the default.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1026,3 +1026,11 @@ Im ℑ
 dotless
   .i ı
   .j ȷ
+
+die
+  .six ⚅
+  .five ⚄
+  .four ⚃
+  .three ⚂
+  .two ⚁
+  .one ⚀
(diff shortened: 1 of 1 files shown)


emilyyyylime:
This seems fine to me but I'd like to hone in on exactly which characters we aim to provide here, which is why I opened #5. That isn't to say I feel this PR is out of scope for the project

mkorje (approved):
Given we've come to some consensus in #5, I believe these additions "could be useful" and are certainly convenient names.

emilyyyylime (approved):

src/modules/sym.txt
```diff
@@ -1026,3 +1026,11 @@ Im ℑ
 dotless
   .i ı
   .j ȷ
+
```
emilyyyylime, line 1029:
```suggestion
+ // Miscellany.
```

emilyyyylime (approved):

PR #9, opened 2024-11-19
Implement parsing Unicode names

### LINKED ISSUE #6: Suggestion: include Unicode codepoint standard names to sym.txt
Many symbols in [`sym.txt`](https://github.com/typst/codex/blob/main/src/modules/sym.txt) are specified as their Unicode codepoint in the form `U+XXXX` rather than a plain character, because it would be hard to parse or notice when reading the file later. I believe using the Unicode-assigned name of such characters would be more useful and self-documenting than simply entering the code point.

Ideally, these names would be machine-checked in `build.rs` rather than just act as informative comments, to ease the minds of reviewers from ensuring the right name is provided for each character. These names could also then be used to look up the wanted Unicode codepoint thereby entirely replacing the `U+` scalar reference.

Either way, we could opt to include the names even on characters that are directly embedded in the txt files just to have more context directly available when editing them (though this is definitely more of a bonus/personal preference change and should be discussed separately.)

emilyyyylime:
This is locked behind an optional build dependency to avoid incurring the cost of `unicode_names2` to every downstream package. (It really is [much heavier than it needs to be](https://github.com/progval/unicode_names2/pull/42) :sweat_smile:)

To reiterate; this PR does not amount to any extra build time or dependencies for downstream packages (unless they explicitly opt in to the `unicode_names2` feature.)

The current syntax consist of simply appending the name after the U+XXXX literal, some alternatives that were discussed:
<ol type="A">
<li> <pre>zwj U+200D "Zero width joiner"</pre> </li>
<li>
  <pre>zwj U+200D // Zero width joiner</pre>
  (possibly also using a specialised comment token like <code>///</code> or <code>//!</code>)
</li>
</ol>

Also as the code is implemented currently, explicit names cannot be added after verbatim characters. This was mostly an arbitrary choice and might be rethought if we decide on implementing one of the other syntax options above.

Another area with some room to behave differently is this being a build dependency, rather than a dev-dependency. This was done because writing the verification as a separate test would require reimplementation of the parsing code from the ground up (dev-dependencies are only available inside tests/benchmarking code).

#### Example output for bad names:
image

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -131,7 +131,7 @@ fn decode_char(text: &str) -> StrResult<char> {
             .and_then(|n| char::from_u32(n))
             .ok_or_else(|| format!("invalid unicode escape {hex:?}"))?;
 
-        #[allow(unused_variables)]
+        #[cfg_attr(not(feature = "unicode_names2"), expect(unused_variables))]
         if let Some(name) = name {
             #[cfg(feature = "unicode_names2")]
             if unicode_names2::character(name) != Some(ch) {
--- a/build.rs
+++ b/build.rs
@@ -86,8 +86,8 @@ fn tokenize(line: &str) -> StrResult<Line> {
         return Ok(Line::Blank);
     }
 
-    let (head, tail) = match line.split_once(' ') {
-        Some((a, b)) => (a, Some(b)),
+    let (head, tail) = match line.split_once(char::is_whitespace) {
+        Some((a, b)) => (a, Some(b.trim_start())),
         None => (line, None),
     };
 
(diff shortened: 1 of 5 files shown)


build.rs
```diff
@@ -121,10 +121,29 @@ fn validate_ident(string: &str) -> StrResult<()> {
...
+        };
+
+        let ch = u32::from_str_radix(hex, 16)
             .ok()
-            .and_then(|n| char::try_from(n).ok())
-            .ok_or_else(|| format!("invalid unicode escape {text:?}"))
+            .and_then(|n| char::from_u32(n))
+            .ok_or_else(|| format!("invalid unicode escape {hex:?}"))?;
+
+        #[allow(unused_variables)]
```
MDLC01, line 134:
```suggestion
        #[expect(unused_variables)]
```
We may want to use `expect` here.

emilyyyylime: In this case `allow` is correct because the variable will not be unused if the feature is enabled. This might be a case for `cfg_attr` though—let me try that

MDLC01: You're right

MDLC01:
This would close #6, right?

emilyyyylime:
Yeah, that's the idea

build.rs
laurmaedje: expect might've also screwed with Typst's MSRV

laurmaedje:
I think it might be a little cleaner to have a new feature `validate-unicode-escapes = ["dep:unicode_names2"]` and use that externally.

MDLC01:
I think it may be more future-proof to force the name to be quoted (alternative A. in OP), as suggested in https://github.com/typst/codex/pull/2#issuecomment-2501946380.

emilyyyylime:
Yeah, I can implement the feature name and quoting + resolve conflicts

MDLC01:
I will close this for now for inactivity. Feel free to reopen if you come back to this in the future.

PR #14, opened 2024-11-29
Remove compatibility symbols (Proposal 19)

MDLC01:
## Description

This PR implements Proposal 19 (~iteration 1~ iteration 2) from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). It removes compatibility codepoints. For some context, see [the Unit Symbols paragraphs of section 22.2.1 of the Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G20445), as well as the description of Proposal 19 in the aforementioned Symbol Proposals document (copied below for simplicity).

Proposal 19 (iteration 2).

`angle.l.double` and `angle.r.double` change codepoints. `degree.c`, `degree.f`, `ohm`, `ohm.inv`, `angstrom`, and `kelvin` become invalid.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -67,11 +67,11 @@ angle ∠
   .l ⟨
   .l.curly ⧼
   .l.dot ⦑
-  .l.double 《
+  .l.double ⟪
   .r ⟩
   .r.curly ⧽
   .r.dot ⦒
-  .r.double 》
+  .r.double ⟫
   .acute ⦟
   .arc ∡
   .arc.rev ⦛
@@ -532,8 +532,6 @@ join ⨝
   .l ⟕
   .l.r ⟗
 degree °
-  .c ℃
-  .f ℉
 smash ⨳
 
 // Currency.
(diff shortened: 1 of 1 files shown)


laurmaedje:
I'm okay with removing the compatibility symbols.

mkorje:
I agree we should add `Omega.inv`, and I'd argue in place of `mho` since I imagine a user might expect `ohm` to also work then. But as `mho` is a Unicode informative alias and I don't know anything about its use, I'm not quite sure if omitting `mho` would be good or bad.

dccsillag:
+1 for `Omega.inv`.

MDLC01:
I replaced `mho` with `Omega.inv`, as "mho" seems to mostly be an outdated unit. Of course, we can always add `mho` later if necessary.

Enivex:
It's so cute though!

mkorje (approved):

T0mstone (approved):

Issue #15, opened 2024-12-10
Support for ≒ U+2252 and ≓ U+2253


sjfhsjfh:
Adding ≒ `eq.fallingdots` and ≓ `eq.risingdots` for U+2252 (Approximately Equal To or the Image Of) and U+2253 (Image of or Approximately Equal To)

In LaTeX package `amssymb` these two are represented by `\fallingdotseq` and `\risingdotseq`.

The former one is also used as an alternative for ≈ (`approx`) in certain countries and regions (e.g. Japan and Taiwan).

PR #16, opened 2024-12-10
Add U+2252 and U+2253

sjfhsjfh:
#15

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -260,12 +260,13 @@ eq =
   .star ≛
   .circle ⊜
   .colon ≕
+  .dots ≑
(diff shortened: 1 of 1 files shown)


MDLC01:
Thanks for the contribution!

Regarding the name, `eq.dots.down` and `eq.dots.up` would probably be more consistent with the other symbols (such as `dots.down` for U+22F1 ⋱ and `dots.up` for U+22F0 ⋰).

MDLC01:
Additionally, ≑ could be `eq.dots`.

MDLC01 (approved):

mkorje (approved):

PR #17, opened 2024-12-12
Add `inter` as an alternative for `sect` (Proposal 14, part I)

MDLC01:
This PR implements the first part of Proposal 14 (iteration 1) from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). It adds `inter` as an alternative for `sect`. Eventually, `sect` should be deprecated, and later removed. For more context, see [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1311370099382292520).

For now, there is no infrastructure to deprecate symbols in Codex, nor is there a way to deprecate a constant in Typst. This makes it impossible to deprecate `sect` immediately. However, since `sect` is a commonly used symbol, I believe the sooner we introduce its new name, the better.

Since this PR assumes the future removal of a symbol (`sect`) and a variant (`integral.sect`), and even though this was already discussed on Discord, cc. @laurmaedje.

PR #19, opened 2024-12-14
Make it possible to deprecate symbols and modules

MDLC01:
This PR adds the ability to deprecate definitions (i.e., symbols and modules), as suggested in https://github.com/typst/codex/pull/2#issuecomment-2501943047.

Syntax-wise, marking a definition as deprecated can be done by adding <code>@deprecated: <i>deprecation message</i></code> to the line preceding the definition.

Currently, this feature is not used, but it could be used to mark  `sym.inter` as deprecated in #17.[^1]

[^1]: Proposal 14 (partially implemented in #17) also removes the `sym.integral.sect` variant, which, as a variant, cannot be deprecated with this PR. This is probably fine in this specific case, because `sym.integral.sect` is not a highly used symbol, but we may want the ability to deprecate specific variants in the future.

Deprecated definitions have an associated message. For now, Typst does not support deprecating constants, but this is implemented in typst/typst#5582.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,19 +5,19 @@ use std::path::Path;
 type StrResult<T> = Result<T, String>;
 
 /// A module of definitions.
-struct Module<'a>(Vec<(&'a str, Entry<'a>)>);
+struct Module<'a>(Vec<(&'a str, Binding<'a>)>);
 
 impl<'a> Module<'a> {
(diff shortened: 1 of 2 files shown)


MDLC01:
Given this is a change to the interface, cc. @laurmaedje.

laurmaedje (approved):

laurmaedje:
I renamed `Entry` to `Binding` for consistency with Typst.

Thank you!

Issue #5, opened 2024-11-17
Formally define the goals of Codex


emilyyyylime:
I think we should have some agreed upon document (even just as part of the README) directly stating which characters are or are not in scope for inclusion in this project, and how simple should a given character be to access.

This would also include some guiding principle for assigning names to characters, such as when abbreviations are okay and when they aren't, when can a character be accessible through multiple distinct names (and whether one of them should be considered "canonical"?), whether we strive to describe the usage of characters, their origin/formal meaning, or their visual appearance (or rather when do we do which).

In my opinion this will greatly help prioritise new additions to the repo and help reviewers to decide what changes to approve

(side note: do we have a preference for how to stylise the name; e.g. codex, Codex, or CodeX?)

dccsillag:
Strong agree.

MDLC01:
Regarding the scope, I would say any Unicode character that is not part of a natural writing system may be considered for inclusion. But this may be too broad, and it is also immediately contradicted by most Greek and Hebrew letters being already included (and rightfully so).

> (Re. name styling: all sound good to me.)

Please not CoDeX! I would say Codex (or CODEX in a capitalize context) is fine.

emilyyyylime:
Are there any specific Unicode (or other) categorisations we could use to help refine this definition? I think it should firstly focus on the needs of Typst users (⇒ symbols used in academic writing and other typesetting settings, that are often not easily accessible in the keyboard layouts of users wishing to type them), and expand from there with more specific examples

mkorje:
I'm of the opinion that Codex's scope should be fairly broad: to assign names to most Unicode characters. Whilst I agree that we should focus on the needs of Typst users first, I wouldn't want this to discourage adding names for symbols that fall outside of this. For example, #3, I think is a fine addition. Die face symbols aren't apart of any natural writing system (afaik...), and the name is clear-cut.

Regardless, I think any scope we set is destined to end up being too broad/restrictive. So I share @MDLC01's view that anything not part of a natural writing system may be considered, and things part of a natural writing system _may_ be allowed if its inclusion makes sense. I'd suggest that the criterion for this inclusion is something along the lines of substantial usage and accessibility in users' keyboard layouts. (And this would rightly justify the inclusion of most Greek and Hebrew characters, I believe.)

With regards to naming guidelines, I think writing out the existing implicit "rules" would be a great start. For example, `.rev` and `.not` being the standard modifiers for the reversed variant of a character and the variant of a character with a forward slash through it.

[...]

PR #20, opened 2024-12-16
Add math styling module

mkorje:
This PR adds functions and the related infrastructure needed to take a character/string and a style in math (e.g. blackboard-bold, monospace), and return it in styled form. It is implemented under the feature `styling`, which I've added as a default.

Moving this to Codex was brought up briefly on Discord, and whilst it doesn't directly fall under "naming Unicode symbols", I think it still fits quite nicely in this crate. I may have jumped the gun a little on this, but it wasn't much work to do. Hopefully this PR starts some more thorough discussion of this idea. I also think that moving this outside of the main typst repo will be an improvement and make it more maintainable.

### DIFF
--- a/src/styling.rs
+++ b/src/styling.rs
@@ -26,6 +26,7 @@ pub const UNICODE_VERSION: (u8, u8, u8) = (16, 0, 0);
 /// - digamma: The uppercase and lowercase digamma, Ϝ (U+03DC) and ϝ (U+03DD).
 /// - dotless: The dotless variants of the lowercase Latin letters i and j, ı
 ///   (U+0131) and ȷ (U+0237).
+/// - hebrew: The Hebrew letters א–ד (U+05D0..U+05D3).
 ///
 /// Note that some styles support only a subset of a group. The characters each
 /// style supports are given in their documentation.
@@ -163,6 +164,10 @@ pub enum MathStyle {
     ///
     /// Supported characters: latin.
     BoldRoundhand,
+    /// Hebrew letterlike math symbols.
+    ///
+    /// Supported characters: hebrew.
+    Hebrew,
 }
 
 /// Base [`MathStyle`]s used in Typst.
@@ -263,6 +268,7 @@ impl MathStyle {
             (_, _, Some(true) | None) if matches!(c, 'ı' | 'ȷ' | 'ħ') => {
                 MathStyle::Italic
             }
+            (_, _, Some(true) | None) if is_hebrew(c) => MathStyle::Hebrew,
             _ => MathStyle::Plain,
         }
     }
@@ -389,6 +395,7 @@ pub fn to_style(c: char, style: MathStyle) -> ToStyle {
         BoldChancery => to_bold_chancery(c),
         Roundhand => to_roundhand(c),
         BoldRoundhand => to_bold_roundhand(c),
+        Hebrew => [to_hebrew(c), '\0'],
     };
     ToStyle::new(styled)
 }
@@ -432,6 +439,11 @@ mod conversions {
     #[inline]
     pub fn is_lower_greek(c: char) -> bool {
         matches!(c, 'α'..='ω' | '∂' | 'ϵ' | 'ϑ' | 'ϰ' | 'ϕ' | 'ϱ' | 'ϖ')
+    }
+
+    #[inline]
+    pub fn is_hebrew(c: char) -> bool {
+        matches!(c, 'א'..='ד')
     }
 
     /// The character given by adding `delta` to the codepoint of `c`.
@@ -912,5 +924,15 @@ mod conversions {
     pub fn to_bold_roundhand(c: char) -> [char; 2] {
         // Variation Sequences (Latin script characters)
         [to_bold_script(c), VARIATION_SELECTOR_2]
+    }
+
+    pub fn to_hebrew(c: char) -> char {
+        let delta = match c {
+            // Letterlike Symbols Block (U+2100..U+214F)
+            // Hebrew letterlike math symbols (U+2135..U+2138)
+            'א'..='ד' => 0x1B65,
+            _ => return c,
+        };
+        apply_delta(c, delta)
     }
 }
--- a/src/styling.rs
+++ b/src/styling.rs
@@ -907,23 +907,27 @@ mod conversions {
     pub fn to_chancery(c: char) -> [char; 2] {
         // Standardized Variation Sequences (uppercase Latin script characters)
         // Variation Sequences (lowercase Latin script characters)
-        [to_script(c), VARIATION_SELECTOR_1]
+        let next = if is_latin(c) { VARIATION_SELECTOR_1 } else { '\0' };
+        [to_script(c), next]
     }
 
     pub fn to_bold_chancery(c: char) -> [char; 2] {
         // Variation Sequences (Latin script characters)
-        [to_bold_script(c), VARIATION_SELECTOR_1]
+        let next = if is_latin(c) { VARIATION_SELECTOR_1 } else { '\0' };
+        [to_bold_script(c), next]
     }
 
     pub fn to_roundhand(c: char) -> [char; 2] {
         // Standardized Variation Sequences (uppercase Latin script characters)
         // Variation Sequences (lowercase Latin script characters)
-        [to_script(c), VARIATION_SELECTOR_2]
+        let next = if is_latin(c) { VARIATION_SELECTOR_2 } else { '\0' };
+        [to_script(c), next]
     }
 
     pub fn to_bold_roundhand(c: char) -> [char; 2] {
         // Variation Sequences (Latin script characters)
-        [to_bold_script(c), VARIATION_SELECTOR_2]
+        let next = if is_latin(c) { VARIATION_SELECTOR_2 } else { '\0' };
+        [to_bold_script(c), next]
     }
 
     pub fn to_hebrew(c: char) -> char {
--- a/src/styling.rs
+++ b/src/styling.rs
@@ -290,6 +290,60 @@ pub fn to_style(c: char, style: MathStyle) -> ToStyle {
     ToStyle::new(styled)
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum MathVariant {
+    Plain,
+    Script,
+    Fraktur,
+    SansSerif,
+    Monospace,
+    DoubleStruck,
+}
+
+pub fn resolve_style(
+    c: char,
+    variant: Option<MathVariant>,
+    bold: bool,
+    italic: Option<bool>,
+) -> MathStyle {
+    use conversions::*;
+    use MathVariant::*;
+    match (variant.unwrap_or(Plain), bold, italic) {
+        (Plain, false, Some(true)) if is_latin(c) | is_greek(c) => MathStyle::Italic,
+        (Plain, true, Some(false)) if is_latin(c) | is_greek(c) => MathStyle::Bold,
+        (Plain, true, Some(true)) if is_latin(c) | is_greek(c) => MathStyle::BoldItalic,
+        (Plain, true, _) if is_digit(c) | matches!(c, 'Ϝ' | 'ϝ') => MathStyle::Bold,
+        (Plain, _, Some(true)) if matches!(c, 'ı' | 'ȷ' | 'ħ') => MathStyle::Italic,
+        (Plain, _, None) => resolve_style(c, variant, bold, Some(is_italic(c))),
+        (Plain, _, _) => MathStyle::Plain,
+
+        (SansSerif, false, Some(false)) if is_latin(c) => MathStyle::SansSerif,
+        (SansSerif, false, Some(true)) if is_latin(c) => MathStyle::SansSerifItalic,
+        (SansSerif, true, Some(false)) if is_latin(c) => MathStyle::SansSerifBold,
+        (SansSerif, true, Some(true)) if is_latin(c) => MathStyle::SansSerifBoldItalic,
+        (SansSerif, false, _) if is_digit(c) => MathStyle::SansSerif,
+        (SansSerif, true, _) if is_digit(c) => MathStyle::SansSerifBold,
+        (SansSerif, _, Some(false)) if is_greek(c) => MathStyle::SansSerifBold,
+        (SansSerif, _, Some(true)) if is_greek(c) => MathStyle::SansSerifBoldItalic,
+        (SansSerif, _, None) => resolve_style(c, variant, bold, Some(is_italic(c))),
+
+        (Script, false, _) if is_latin(c) => MathStyle::Script,
+        (Script, true, _) if is_latin(c) => MathStyle::BoldScript,
+        (Fraktur, false, _) if is_latin(c) => MathStyle::Fraktur,
+        (Fraktur, true, _) if is_latin(c) => MathStyle::BoldFraktur,
+        (Monospace, _, _) if is_digit(c) | is_latin(c) => MathStyle::Monospace,
+        (DoubleStruck, _, Some(true)) if matches!(c, 'D' | 'd' | 'e' | 'i' | 'j') => {
+            MathStyle::DoubleStruckItalic
+        }
+        (DoubleStruck, _, _)
+            if is_digit(c) | is_latin(c) | matches!(c, '∑' | 'Γ' | 'Π' | 'γ' | 'π') =>
+        {
+            MathStyle::DoubleStruck
+        }
+        (_, _, _) => resolve_style(c, None, bold, italic),
+    }
+}
+
 /// Functions which convert a `char` to its specified styled form.
 ///
 /// Sourced from:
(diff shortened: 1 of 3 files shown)


MDLC01:
A bit more context: https://github.com/typst/codex/issues/5#issuecomment-2488669690.

src/styling.rs
```diff
@@ -0,0 +1,613 @@
...
+        styled
+    }
+}
+
+/// Functions which map a [`char`] to its specified styled form.
+///
+/// Sourced from:
+/// - [Mathematical Alphanumeric Symbols]
+/// - [Arabic Mathematical Alphabetic Symbols]
+/// - <https://www.w3.org/TR/mathml-core/#new-text-transform-mappings>
```
MDLC01, line 160: This should probably link the Unicode Core Specification as well. For example, <https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G52612>.

src/styling.rs
```diff
@@ -0,0 +1,613 @@
...
+            'α'..='ω' => 0x1D311,
+            '∂' => 0x1B4D9,
+            'ϵ' => 0x1D2E7,
+            'ϑ' => 0x1D30C,
+            'ϰ' => 0x1D2EE,
+            'ϕ' => 0x1D30A,
+            'ϱ' => 0x1D2EF,
+            'ϖ' => 0x1D30B,
+            'Ϝ' | 'ϝ' => 0x1D3EE,
+            '0'..='9' => 0x1D79E,
```
MDLC01, line 216: I think it would be better to somehow use characters (e.g., `'𝐀'`) instead of codepoints on the right hand sides here.

src/styling.rs
```diff
@@ -0,0 +1,613 @@
...
+        apply_delta(c, delta)
+    }
+
+    /// To chancery style of script symbols.
+    ///
+    /// This function returns a variation sequence using the variation selector
+    /// `U+FE00`, as specified in the [`StandardizedVariants.txt`] file from
+    /// the Unicode Character Database. Note that only the capital latin
+    /// letters are standardized variation sequences, but this function also
+    /// maps the small latin letters.
```
MDLC01, line 407:
```suggestion
    /// the Unicode Character Database. Note that only the capital Latin
    /// letters are standardized variation sequences, but this function also
    /// maps the small Latin letters.
```

And similar in three other places.

MDLC01:
Regardless of my above comments, @laurmaedje will have to approve moving those functions before we merge this PR.

mkorje:
> Regarding the implementation, as well as having many `to_{style}` functions, there could also be a `MathStyle::apply` method to apply a style to a `char`.

The issue I had with something like this is that some of the styles will output multiple characters.

I also considered implementing the `MathStyling` trait for `EcoString`, as `EcoString` is used in the Typst repo. Though I wasn't sure about adding that as a dependency; maybe instead it could be under a feature flag?

src/styling.rs
mkorje: At the moment the right hand side is just a delta to add to the original codepoint. I can't quite remember why exactly I did it this way instead of just mapping directly to the new codepoint, in which case we could use characters instead.

T0mstone: The delta allows the pattern to be `'A'..='Z'` instead of 26 individual ones.

mkorje: Oh of course, for some reason I thought `'A'..='Z' => '𝐀'..='𝐙'` would work :sweat_smile:.

MDLC01:
> > Regarding the implementation, as well as having many `to_{style}` functions, there could also be a `MathStyle::apply` method to apply a style to a `char`.
> 
> The issue I had with something like this is that some of the styles will output multiple characters.

Can you not return `[char; 2]` like the `to_style` function? Essentially, making `to_style` a method on `MathStyle`.

> I also considered implementing the `MathStyling` trait for `EcoString`, as `EcoString` is used in the Typst repo. Though I wasn't sure about adding that as a dependency; maybe instead it could be under a feature flag?

Instead, maybe `to_styled` could be made generic over its return type `T: Default + AddAssign<&str>`. Or maybe I'm just overthinking, and we can just add the dependency behind a feature flag.

MDLC01:
Also, you can implement `MathStyling` for `str` instead, that way anything that derefs to `str` (such as `String` or `EcoString`) can benefit from `to_styled`.

laurmaedje:
Looks like there's quite a bit more going on here than in typst-layout/src/math/text.rs, e.g. the looped and chancery stuff. Would that somehow be exposed in Typst or is it for completeness?

laurmaedje:
Regarding EcoString: I wouldn't add a dependency on it here. Doesn't seem worth it to me.

src/styling.rs
```diff
@@ -0,0 +1,616 @@
+/*!
+Style mathematical symbols in Unicode.
+*/
```
laurmaedje, line 3:
```suggestion
//! Style mathematical symbols in Unicode.
```

src/styling.rs
```diff
@@ -0,0 +1,616 @@
...
+///
+/// ```
+/// use codex::styling::{to_style, MathStyle};
+///
+/// assert_eq!(['𝑴', '\0'], to_style('M', MathStyle::SerifItalicBold));
+/// assert_eq!(['𞸪', '\0'], to_style('ك', MathStyle::Initial));
+/// assert_eq!(['ⅈ', '\0'], to_style('i', MathStyle::DoubleStruckItalic));
+/// assert_eq!(['𝓴', '\u{fe01}'], to_style('k', MathStyle::RoundhandBold));
+/// ```
+pub fn to_style(c: char, style: MathStyle) -> [char; 2] {
```
laurmaedje, line 77: The `[char: 2]` approach where `\0` has special meaning seems a bit easy to misuse to me. Maybe it could return a custom iterator type instead? Then to string impl would be trivial to do yourself (`s.chars().flat_map(|c| to_style(c, style)).collect()`. I would probably even remove the trait then.

mkorje:
> Looks like there's quite a bit more going on here than in typst-layout/src/math/text.rs, e.g. the looped and chancery stuff. Would that somehow be exposed in Typst or is it for completeness?

Yes, I plan to expose these in Typst as well. At this point in time it is just for completeness though.

For the Arabic ones I plan to add to [Variants](https://typst.app/docs/reference/math/variants) the functions `inital`, `tailed`, `looped`, and `stretched`, as well as `isolated` as an alias for `serif`. Though I should note I have no idea if these should even really have their names in English, or if shortened names like `tail` or `loop` would work.

As for the chancery/roundhand stuff, we'd add `scr` to join `cal` in Typst. But there is still some work to do before it can be added (cf. [typst/typst 5853](https://github.com/typst/typst/issues/5853)).

src/styling.rs
mkorje: I agree an iterator is much nicer. I've basically mimicked [`to_uppercase`](https://doc.rust-lang.org/std/primitive.char.html#method.to_uppercase) and [`ToUppercase`](https://doc.rust-lang.org/std/char/struct.ToUppercase.html) from `std`. I also got rid of the trait.

mkorje:
I've changed the `to_style` function to return a custom iterator type and have removed the `MathStyling` trait for now (see https://github.com/typst/codex/pull/20#discussion_r1966739628)

> Instead, maybe `to_styled` could be made generic over its return type `T: Default + AddAssign<&str>`. Or maybe I'm just overthinking, and we can just add the dependency behind a feature flag.

> Also, you can implement `MathStyling` for `str` instead, that way anything that derefs to `str` (such as `String` or `EcoString`) can benefit from `to_styled`.

Since I've removed the trait for now, this no longer applies.

> Regarding the implementation, as well as having many `to_{style}` functions, there could also be a `MathStyle::apply` method to apply a style to a `char`.

I think in my previous reply to this I misunderstood what you were saying, sorry. I've done that now.

Also, I was wondering whether to make the `to_style` function a method on `char` instead?

I also need to double check Arabic, as for the normal style the math symbols have separate codepoints (unlike latin where the upright serif for math is the same as the standard codepoints), I think.

laurmaedje:
I agree that we shouldn't have `[char; 2]` in the public API. I don't think we need a public `MathStyle::apply`. I think we can inline its implementation into `to_style`.

mkorje:
When you say not having `[char; 2]` in the public API, does that include the functions in `mappings`?

mkorje:
Ok I've redone some stuff and more thoroughly documented/commented everything, which makes the code all much clearer I think. I also double checked the deltas and cross referenced with MathML Core. I think this ready from my end to review.

> I hadn't seen that those are public. Feels slightly duplicate to me with the enum. It's nice to have have in the type that they return just one char. I guess it depends a bit on how you plan to use them in Typst.

I just decided to make those functions private in the end.

mkorje:
Oh there is one thing left: we should maybe (?) switch alef, bet, dalet, gimel to map to their mathematical glyphs instead of the Hebrew letters.

src/styling.rs
```diff
@@ -0,0 +1,788 @@
+//! Style mathematical symbols in Unicode.
+
+use std::fmt::{self, Write};
+use std::iter::FusedIterator;
+
+/// The version of [Unicode](https://www.unicode.org/) that this version of the
```
MDLC01, line 6:
```suggestion
/// The version of [Unicode](https://www.unicode.org/versions/Unicode16.0.0/core-spec/) that this version of the
```

This may be a more useful link.

src/styling.rs
```diff
@@ -0,0 +1,788 @@
...
+/// styling module is based on.
+pub const UNICODE_VERSION: (u8, u8, u8) = (16, 0, 0);
+
+/// A style for mathematical symbols.
+///
+/// Notation in mathematics uses a basic set of characters which can be styled.
+/// The following groupings are used in the documentation:
+/// - digits: The basic Latin digits 0–9 (U+0030..U+0039).
+/// - latin: The basic uppercase and lowercase Latin letters, a–z
```
MDLC01, line 15:
```suggestion
/// - Latin: The basic uppercase and lowercase Latin letters, a–z
```

"Latin" should be capitalized, same for "Greek" and "Arabic" below.

src/styling.rs
```diff
@@ -0,0 +1,788 @@
...
+    Bold,
+    /// Italic style. May be serif or sans-serif depending on the font.
+    ///
+    /// Supported characters: latin, greek, dotless, and the extra ħ (U+0127).
+    Italic,
+    /// Bold italic style. May be serif or sans-serif depending on the font.
+    ///
+    /// Supported characters: latin, greek.
+    BoldItalic,
+    /// Script style.
```
MDLC01, line 75: May be chancery or roundhand depending on the font.

src/styling.rs
```diff
@@ -0,0 +1,788 @@
...
+    /// Supported characters: latin.
+    Fraktur,
+    /// Bold fraktur style. Also known as bold black-letter style.
+    ///
+    /// Supported characters: latin.
+    BoldFraktur,
+    /// Sans-serif style.
+    ///
+    /// Supported characters: digits, latin.
+    Sansserif,
```
MDLC01, line 94:
```suggestion
    SansSerif,
```

Not very important, but writing "sans serif" as one word feels weird to me. Another alternative is simply `Sans`.

T0mstone: `Sans` is very familiar since it's present in tons of font names (e.g. Comic Sans), so I'd slightly prefer that, especially given the length of `SansBoldItalic` vs `SansSerifBoldItalic`.

src/styling.rs
```diff
@@ -0,0 +1,788 @@
...
+    ///
+    /// Supported characters: arabic excluding ك (U+0643).
+    Looped,
+    /// Double-struck style. Also known as open-face style or blackboard-bold
+    /// style.
+    ///
+    /// Supported characters: digits, latin, arabic excluding ا (U+0627),
+    /// ك (U+0643), ه (U+0647), and the extras ∑ (U+2211), Γ (U+0393), Π
+    /// (U+03A0), γ (U+03B3), π (U+03C0).
+    Doublestruck,
```
T0mstone, line 143: Same as with `Sansserif` here.
Either
```suggestion
 DoubleStruck,
```
or I'd also propose just using
```suggestion
 Double,
```

MDLC01: Simply "double" instead of "double struck" isn't nearly as common as "sans" for "sans serif", so I think we should keep "double struck" here.

src/styling.rs
mkorje: Good catch, will add that.

src/styling.rs
mkorje: Though that would then need updating with each version change as well. Not sure if this is really much of a point though.

src/styling.rs
mkorje: The name used is double-struck in Unicode, so I figure its best to stick to that.

src/styling.rs
mkorje: IIRC my thought was that hyphens shouldn't be camel case otherwise it reads like two separate words "sans serif" instead of "sans-serif".

As with double-struck, despite being more verbose I think it'd be good to keep with the full name "sans-serif" used in Unicode.

laurmaedje: I think `SansSerif` is definitely more readable than `Sansserif`

src/styling.rs
MDLC01: Well, the tuple below will have to be updated as well anyway. But alternatively you can replace `Unicode16.0.0` with `latest`.

src/styling.rs
T0mstone: > IIRC my thought was that hyphens shouldn't be camel case otherwise it reads like two separate words "sans serif" instead of "sans-serif".

It *is* two separate words. Sans means "without", so it's just saying "without serifs".

src/styling.rs
mkorje: That could also cause issues as `latest` might not match the version here :sweat_smile:

src/styling.rs
mkorje: I think I meant that as two separate words they could then each read like a style modifier (e.g. Script and BoldScript): SansSerifBold is Sans-serif Bold not Sans (Serif Bold). Anyways I just made the change to SansSerif, and also to DoubleStruck.

Enivex:
What is the status of this exactly?

mkorje:
I need to open a PR on the Typst side. I think I almost finished it a while ago, I'll finish it soon...

mkorje:
Ok, this should all be done for the codex side of things. See https://github.com/typst/typst/pull/6309 for the accompanying Typst PR.

The last commit moves the `MathVariant` enum from Typst here and adds some logic with the `resolve_style` function to decide on the style to use given options for the variant, bold, and italic (and improves a bit the user experience in Typst - more in the PR linked above). You'll notice that the Arabic variants and roundhand/chancery are missing from the `MathVariant` enum here. I'll only add them there once they're ready to be used in Typst, which isn't the case at the moment (and I didn't want that to block this PR). I guess when that does get updated it'd maybe warrant a minor version bump?

src/styling.rs
```diff
@@ -0,0 +1,867 @@
...
+        apply_delta(c, delta)
+    }
+
+    pub fn to_script(c: char) -> char {
+        let delta = match c {
+            // Letterlike Symbols Block (U+2100..U+214F)
+            // Letterlike symbols (U+2100..U+2134)
+            'g' => 0x20A3,
+            'H' => 0x20C3,
+            'I' => 0x20C7,
```
laurmaedje, line 487: are these deltas are computed by hand? I only now noticed that these numbers aren't codepoint numbers as in the previous code but rather deltas.

mkorje: I think it's just one of the Arabic styles that was completely new and done by hand

laurmaedje: Ah, the spec also defines it in terms of deltas?

mkorje: It's technically not in the spec, but in the non-normative [C. Mathematical Alphanumeric Symbols](https://www.w3.org/TR/mathml-core/#new-text-transform-mappings). I have no idea where they have all gone, suddenly there's only italic there now. Anyways, it had both the original codepoint, new one, and the delta.

Delta just ends up being a bit nicer as there tend to be chunks that all have the same delta.

src/styling.rs
```diff
@@ -0,0 +1,867 @@
...
+pub enum MathVariant {
+    Plain,
+    Script,
+    Fraktur,
+    SansSerif,
+    Monospace,
+    DoubleStruck,
+}
+
+pub fn resolve_style(
```
laurmaedje, line 303: I think this might be nicer as a method `MathStyle::select`. A bit of docs would also be great, especially for the `None` cases and on the fallback behaviour.

src/styling.rs
```diff
@@ -0,0 +1,867 @@
...
+        DoubleStruckItalic => [to_double_struck_italic(c), '\0'],
+        Chancery => to_chancery(c),
+        BoldChancery => to_bold_chancery(c),
+        Roundhand => to_roundhand(c),
+        BoldRoundhand => to_bold_roundhand(c),
+    };
+    ToStyle::new(styled)
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
```
laurmaedje, line 293: Could you add one sentence of docs here as well, just so that all public items are documented?

src/styling.rs
mkorje: Can do.
Though I was a bit unsure here with docs and all since it's basically going to just follow Typst. And so at the moment the new enum will be missing some of the styles initially.

mkorje: The MathVariant one you left the other comment on.

laurmaedje: I guess this piece of the code could also stay in Typst.

mkorje: Though I'd like (and think it'd be nice) to not have any character matching code stuff on the Typst side

I'll leave out the Arabic styles from this enum for now (will add them in a later PR) and add the `non_exhaustive` attribute.

laurmaedje: Since we've reframed the goal of codex to be more generally anything Typst i18n and Unicode related, I'm more fine with having things here be not totally generalized.

mkorje:
I'm blocked by https://github.com/typst/typst/pull/6441 now on the Typst side as https://github.com/typst/codex/pull/86 is causing errors... I'm too tired to try sort this out tonight and I am like 75% confident this should be working on the Typst :laughing:

mkorje:
I should also remove Script from the MathVariants enum as Roundhand and Chancery are already on it.

MDLC01:
Except script is different because it doesn't specify a variation selector, right? If so, we can probably keep it (even if we don't use it in Typst).

mkorje:
Yes, but if the font doesn't support it it just gets ignored.

I thought to remove it because we'd then not be using it in Typst, though I guess there's no harm in keeping it.

src/styling.rs
```diff
@@ -0,0 +1,942 @@
...
+    /// ```
+    /// # use codex::styling::{MathStyle, MathVariant};
+    /// assert_eq!(
+    ///     MathStyle::DoubleStruck,
+    ///     MathStyle::select('R', Some(MathVariant::DoubleStruck), true, Some(true))
+    /// );
+    /// ```
+    pub fn select(
+        c: char,
+        variant: Option<MathVariant>,
```
laurmaedje, line 227: Why is this an `Option` if `Plain` and `None` behave equivalently?

PR #24, opened 2025-01-04
Make `planck` more useful (Proposal 26)

MDLC01:
## Description

This PR implements Proposal 26 (iteration 1) from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). It changes `planck` to ħ (which gets italicized to ℏ in math thanks to https://github.com/typst/typst/pull/5492), and removes the `planck.reduce` variant. For more context, see [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1325065616444624956).

There is no way to take advantage of the symbol deprecation mechanism introduced in #19 to make this change smoother.

`planck.reduce` becomes invalid, and `planck` changes codepoint.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1025,8 +1025,7 @@ ZZ ℤ
 
 // Miscellaneous letter-likes.
 ell ℓ
-planck ℎ
-  .reduce ℏ
+planck ħ
 angstrom Å
 kelvin K
 Re ℜ


mkorje (approved):

T0mstone:
I don't like how this changes people's documents without warning.

I'm still a bit conflicted on whether that's just a necessary evil or whether we should look into adding some kind of ad-hoc warning in typst before making this change here...

MDLC01:
I guess we could split this into multiple phases if we really want to make sure it does not break people's documents:
- First, mark `planck` (but not `planck.reduce` as deprecated (people should use a regular h instead).
- In the next version, change `planck` to be the version with the bar, and deprecated `planck.reduce` (bare `planck` should be used instead).
- Finally, in a third version, remove `planck.reduce`.

I'm not sure it would be worth spanning this over three versions, but if it breaks many documents, I think this is the only way to make a progressive change.

T0mstone:
Hmm..... It may be a bit dumb, but that's probably the most robust way to go about it.

The more I think about it, the more I actually like the idea.

@laurmaedje would you say this is warranted or is it overkill?
I can also accept the viewpoint that doing this change in one go is just something users will have to deal with (tho I don't really like it).

laurmaedje:
Very similar to what I proposed here for `read`: https://github.com/typst/typst/issues/5672

For `read`, it's a bit more critical than for `planck`, but I could imagine going about it that way here as well.

MDLC01:
All this would have to be done on Typst's side, tough. Codex does not have the ability to express such precise warnings.

T0mstone:
After another brief discussion of this [on discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1343989137765830698), I'll drop my concern for this, so LGTM.

T0mstone (approved):

PR #26, opened 2025-01-04
Add descriptive names for `and` and `or`

T0mstone:
Currently, ∧ and ∨ are only addressable as `and` and `or`.

Though that usage is very common, these symbols have various other uses in different fields of math.

Two that I know of are:
- exterior and regressive product in geometry
- infimum and supremum in order theory

Therefore, like `xor` with `plus.circle`, I think these deserve a name that reflects their shape.

I took the names from LaTeX, though I'm not too happy about "vee". Maybe having the same names as LaTeX here is a benefit in and of itself, but I'd also like to hear some other suggestions if you have any.

Side-note: I didn't know where to put them in the source file, so I just went with "Shapes". Feel free to offer a better suggestion.


MDLC01:
They are also used for GCD and LCM in France.

> like `xor` with `plus.circle`

Wouldn't that be like `plus.circle` for `xor`?

> I took the names from LaTeX, though I'm not too happy about "vee".

We already have `ell` for ℓ, and are considering adding `pee` for ℘. With this in mind, I agree `vee` is not a very good name because I would expect it to be a somewhat rounded "v."

dccsillag:
+1.

They are also more generally used for meets and joins in a lattice (related to the order theory example given by @T0mstone).

Also agree that `vee` is unsatisfying, but I don't have any suggestions at the moment.

MDLC01:
@T0mstone note that you would also need to add `sect.wedge` for `sect.and`, and `union.vee` for `union.or`.

> Also agree that `vee` is unsatisfying, but I don't have any suggestions at the moment.

To be fair, I don't really like `wedge` either (although this is more a matter of taste). But at least it probably cannot be confused with anything else, and is well known by people coming from LaTeX.

An alternative to `vee` that I don't really like, but feels worth mentioning, is `wedge.inv`. Or `wedge.t` and `wedge.b`, with `wedge.t` being the default.

laurmaedje:
At first glance, I'm not a big fan of this. Let's discus it more before any merge. (Next week.)

T0mstone:
I've added `sect.wedge` and `union.vee` for now.

As for the discussion, I personally think this is a very reasonable request and if we don't have this, then we should take away `xor`, which I judge as about the same level of niche as the applications mentioned above.

laurmaedje:
What I mainly dislike about this is the following: For the meaning of "and" and "or", `and` and `or` are, in my opinion, much better than `wedge` and `vee`. However, by introducing these two aliases, people coming from LaTeX will just stay with wedge and vee because it's what they are used to. In an attempt of making math markup more semantically correct, we'd likely reduce the average semantic correctness of documents in the wild. This is of course a more general problem, but it's especially pronounced here.

People that use these operators for other meanings than "and" and "or" should probably define their own binding. I can agree that defining `let meet = sym.and` is a bit less clean than `let meet = sym.wedge` (though there's always `let meet = "\u{2227}"`), but I'm not sure that outweighs the downside of hurting the visibility of `and` and `or`.

Enivex:
And an or are semantic names, but they're not the only meanings of these symbols, which can be unfortunate.

That being said, I don't have strong feelings either way

dccsillag:
I agree with @laurmaedje's point.

However, the use of these symbols for meets and joins are also exceptionally common, so I'm inclined to say that we should perhaps also have them as `meet` and `join` -- in fact, these terms subsume their use in logic (and can be seen as a meet, and or can be seen as a join).

I'm unfamiliar with the original motivating example of exterior and regressive products, but a quick search seems to reveal that they are also special cases of meets and joins in the sense of lattices (and are even sometimes called meet and join). So I think my current vote right now is to just also include these as `meet` and `join`, and avoid 'non-descriptive' names (though wedge does actually seem to be occasionally used in geometric algebra).

T0mstone:
If "people coming from LaTeX" is the issue, how about just using different names? Then, if people ask, we can direct them to `and` and `or` first.

For example, they could be `angle.t` and `angle.b`.

laurmaedje:
There is a problem because `join` already exists as `⨝`

MDLC01:
Some more context about the current `join` symbol and bowties, from the [Symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT).

image

In summary, the difference between bowties and the current `join` symbol is the math class, and (roughly) there is not enough symbols in Unicode for it to make sense to have two different symbols (`bowtie` and `join`) with their variants. If we really want to use `join` for the logical or symbol, the current `join` could be renamed to `bowtie.big`.

T0mstone:
Converting this to a draft since it would ideally be done with the changes from #27 instead.

PR #27, opened 2025-01-20
Symbol Aliases

T0mstone:
This PR implements a new framework for defining aliases, i.e. symbols that copy other symbols, and uses that for all of the previously duplicated symbols in `sym`.

We don't intend to have a lot of these, but for the ones we do have, having a system like this is a big improvement to maintainability.
As a good illustration, this PR automatically adds the `arrow` variant from `plus.circle` to `xor`, which seems to have previously been overlooked.

## Guide to creating Aliases
There are two types of aliases:
- Shallow aliases copy a symbol/variant without any other variants. This is the equivalent of `let alias = symbol(a.b.c)` in typst.
- Deep aliases copy a symbol, possibly with some pre-applied modifiers, and all further modifiers. This is the equivalent of `let alias = a.b.c` in typst.

A shallow alias is declared with
```
@<alias> <other symbol name, possibly with modifiers>
```
A deep alias is the same, but additionally ends with `.*`.
Note that this isn't quite like a glob pattern, since e.g. `a.b.*` also matches `a.b` itself.

For some examples, see the diff of `sym.txt`.

## About the Implementation
Properly handling aliases like `@top tack.b` is a bit difficult since modifiers are commutative.
The implementation I wrote uses a fast-path prefix check first and then, if that fails, a slower and heavily allocating algorithm that goes through all modifiers one-by-one.
That way, an alias `@a b.d` for a symbol `b` with a variant `.c.d` correctly copies that to `a.c`.

For the current symbols, the fast path would have been enough, but I didn't want to sacrifice correctness; This could otherwise come back to haunt us later.\
And since the number of aliases will most likely be staying quite low, such a slow and inefficient fallback should be fine.

## Shallow vs Deep Aliases in `sym`
The rules by which I decided the kind of alias for the pre-existing duplicates are basically the following:
- Does the duplicate include all variants? Then it is a deep alias.
- Otherwise, does one copy have variants and the other doesn't? Then the other is a shallow alias.
- Otherwise, does it seem like there will never be any variants? (e.g. the Hebrew letters) Then also a shallow alias.
- Otherwise, it is a deep alias. (see `nabla`) This is to stay open to the possibility of adding variants to it later.

Of course, as mentioned above, `xor` is a bit of an exception since it was missing one of the variants of `plus.circle`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::fmt::Write;
 use std::iter::Peekable;
 use std::path::Path;
@@ -24,6 +25,7 @@ enum Def<'a> {
 enum Symbol<'a> {
     Single(char),
     Multi(Vec<(&'a str, char)>),
+    MultiAlias(Vec<(Cow<'a, str>, char)>),
 }
 
 /// A single line during parsing.
@@ -34,6 +36,7 @@ enum Line<'a> {
     ModuleEnd,
     Symbol(&'a str, Option<char>),
     Variant(&'a str, char),
+    Alias(&'a str, &'a str, &'a str, bool),
 }
 
 fn main() {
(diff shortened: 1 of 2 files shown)


jenuori:
فيس

MDLC01:
This was [discussed on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1336036127055941814) and there was no consensus on whether an alias system is desirable. I'm not sure what we should do with this PR for now. We can probably wait until Typst 0.13.0 is released to re-open the discussion and finally decide whether we want this for now.

T0mstone:
Note that if/when #62 is accepted, `compose` couldn't be an alias like it is here currently (and we'd probably deprecate `circle.stroked.tiny` anyway in a later PR)

T0mstone:
I reverted `mapsto` from being an alias since that would also have picked up `mapsto.double` for `arrow.r.bar.double`. Not sure if we'd want that.

PR #28, opened 2025-01-21
Add more stars

### LINKED ISSUE #25: Emoji Variation selectors

- `sym.hash`, `emoji.hash`
- `sym.trademark`, `emoji.tm`

Prior work: #21, #23.

T0mstone:
I'm happy to receive any feedback about the names.

# Unicode Weirdness
## Four-pointed Stars
There's two black four-pointed stars,
- U+2726 "Black Four Pointed Star"
- U+1F7C6 "Four Pointed Black Star"
I chose to go with the second one since the first one is from the "Dingbats" block, not from "Geometric Shapes Extended".
There's also U+2727 "White Four Pointed Star", which I haven't added since it'd be weirdly inconsistent.

## Five-pointed Stars
I'm _assuming_ the original star is intended to be five-pointed (but idk how much of this is just fonts agreeing with each other rather than Unicode actually saying something about it. Where do you get all that secret Unicode information about how characters are supposed to look?)

In particular, note how there's U+1F7C9 "Light Five Pointed Black Star" and U+1F7CA "Heavy Five Pointed Black Star", but no plain "Five Pointed Black Star".
There's also U+066D "Arabic Five Pointed Star", which I don't know what to do with.

Currently, I've added the former two as `star.filled.{light, heavy}`.

## Six-pointed Stars
At least in my font, U+1F52F "Six Pointed Star with Middle Dot" has an emoji style as default. VS15 does work and we should probably only add this one after having resolved #25, but I've added it for now anyway.

## Eight-pointed Stars
U+1F7D0 is called "Very Heavy Eight Pointed Black Star". What do we call this modifier? I've gone with `veryheavy` for now, but I don't think this is a good choice.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -713,7 +713,51 @@ parallelogram
 star
   .op ⋆
   .stroked ☆
+  .stroked.small ⭒
+  .stroked.outlined ⚝
+  .stroked.shadowed ✰
+  .stroked.filled.l ⯪
+  .stroked.filled.r ⯫
   .filled ★
+  .filled.small ⭑
+  .filled.light 🟉
+  .filled.heavy 🟊
+  .filled.l ⯨
+  .filled.r ⯩
+  .filled.outlined ✭
+  .filled.outlined.heavy ✮
+  .pinwheel ✯
+  .three.filled 🟂
+  .three.filled.medium 🟁
+  .three.filled.light 🟀
+  .three.pinwheel 🟃
+  .four.filled 🟆
+  .four.filled.medium 🟅
+  .four.filled.light 🟄
+  .four.pinwheel 🟇
+  .four.pinwheel.rev 🟈
+  .penta ⛤
+  .penta.inv ⛧
+  .penta.laced.l ⛦
+  .penta.laced.r ⛥
+  .six.filled ✶
+  .six.filled.medium 🟋
+  .six.filled.heavy 🟌
+  .six.pinwheel 🟍
+  .six.dot 🔯
+  .six.david ✡
+  .eight.filled ✴
+  .eight.filled.medium 🟎
+  .eight.filled.heavy 🟏
+  .eight.filled.veryheavy 🟐
+  .eight.filled.rect ✷
+  .eight.filled.rect.heavy ✸
+  .eight.pinwheel ✵
+  .eight.pinwheel.heavy 🟑
+  .twelve ✹
+  .twelve.light 🟒
+  .twelve.heavy 🟓
+  .twelve.heavy.pinwheel 🟔
 
 // Arrows, harpoons, and tacks.
 arrow


T0mstone:
Should I also add the "asterisk" symbols? (e.g. U+1F7AF 🞯 "LIGHT FIVE SPOKED ASTERISK") They're not really "punctuation" like the current `sym.ast`, but they're not officially "stars"...

src/modules/sym.txt
```diff
@@ -713,7 +713,52 @@ parallelogram
 star
   .op ⋆
   .stroked ☆
+  .stroked.small ⭒
+  .stroked.outlined ⚝
```
MDLC01, line 717: I don't really like `.outlined`, but I don't have a good alternative either.

src/modules/sym.txt
```diff
@@ -713,7 +713,52 @@ parallelogram
...
+  .three.pinwheel 🟃
+  .four.filled 🟆
+  .four.filled.medium 🟅
+  .four.filled.light 🟄
+  .four.pinwheel 🟇
+  .four.pinwheel.rev 🟈
+  .penta ⛤
+  .penta.inv ⛧
+  .penta.laced.l ⛦
+  .penta.laced.r ⛥
```
MDLC01, line 742: I don't really like `.laced` here either. This should probably be the same modifier as `.outlined` in my above comment.

MDLC01, line 742:
```suggestion
  .five ⛤
  .five.inv ⛧
  .five.laced.l ⛦
  .five.laced.r ⛥
```

Why not `.five`?

src/modules/sym.txt
```diff
@@ -713,7 +713,52 @@ parallelogram
...
+  .six.filled ✶
+  .six.filled.medium 🟋
+  .six.filled.heavy 🟌
+  .six.pinwheel 🟍
+  .six.david ✡
+  .six.david.dot 🔯
```
MDLC01, line 748:
```suggestion
  .six.stroked ✡
  .six.stroked.dot 🔯
```

Why not `.stroked`?

Alternatively:
```suggestion
  .david ✡
  .david.dot 🔯
```

src/modules/sym.txt
```diff
@@ -713,7 +713,52 @@ parallelogram
...
+  .eight.filled ✴
+  .eight.filled.medium 🟎
+  .eight.filled.heavy 🟏
+  .eight.filled.veryheavy 🟐
```
MDLC01, line 752:
```suggestion
  .eight.filled.light ✴
  .eight.filled 🟎
  .eight.filled.medium 🟏
  .eight.filled.heavy 🟐
```

`.veryheavy` isn't pretty.

MDLC01 (commented):
I left some thoughts in the review comments.

Additionally, maybe U+2726 ✦ should be added as `star.cusp`, `star.filled.cusp`, or `star.four.filled.cusp`.

Moreover, I'm not sure I understand why you said that "attaching them without an extra qualifier is correct" given that the charts mention, as you noted, that five pointed stars constitute a set containing U+1F7C9, U+2605, U+1F7CA, and U+272F.

src/modules/sym.txt
T0mstone: I thought the same too, but the name "outlined" alone doesn't strictly require it to be the same kind of intertwined knot-like design. That might just be a stylistic choice of the font. Maybe Unicode has a guideline somewhere that says how this character should look, but I couldn't find anything.

T0mstone: The other stars don't have strokes through their interior. This one is also just called "pentagram" in unicode, with no mention of "star", but I figured it belonged here. Though now I have a different idea that I'll explain in a subsequent comment.

src/modules/sym.txt
T0mstone: Yeah, I agree, but I don't know if I like "lying" about the weight

T0mstone:
> Moreover, I'm not sure I understand why you said that "attaching them without an extra qualifier is correct" given that the charts mention, as you noted, that five pointed stars constitute a set containing U+1F7C9, U+2605, U+1F7CA, and U+272F.

Note how U+2605 is just `star.filled`. Another idea would be renaming that to `star.five.filled`, then the other ones could follow suit. In that case, we could also rename the pentagram to `star.five.something`, tho I don't really like `star.five.stroked` since the strokes cross each other rather than being an outline only.

T0mstone:
> Additionally, maybe U+2726 ✦ should be added as `star.cusp`, `star.filled.cusp`, or `star.four.filled.cusp`.

In that case we should also add all the other cusps. There are more of them and I was thinking of leaving them for later.

T0mstone:
There, I've added the `.five` modifier to all the regular stars (except for `.op`, which I feel is supposed to be more abstract), since the regular `star.filled` is (by association) five-pointed, and thus (tho I'm not perfectly sure about this) all its plain derivatives can be assumed to be five-pointed as well.

I then went and made the david star and the pentagram the same modifier, namely `.crossing`, since the strokes are crossing each other.

src/modules/sym.txt
MDLC01: Weights just need to be consistent relative to each other. The words by themselves don't carry any meaning I think.

T0mstone: I could imagine the weights to be supposed to be consistent across the different-pointed stars.
But the eight-pointed rectilinear and pinwheel stars seem to be aligned with what unicode calls the medium one in my font, so it's probably safe enough to just do it like you suggested here.

src/modules/sym.txt
MDLC01: A pentagram is indeed exactly that.[^1] So it makes sense for it not to be under `.five`. Maybe it shouldn't even be under `star`, and instead be its own symbol. But I think it's fine the way it is right now.

[^1]: https://en.wikipedia.org/wiki/Pentagram

MDLC01:
Forgive me if I'm missing something obvious: I find it very hard to review so many interconnected symbols. Anyway, here are my comments.

- I still don't like the `.outlined` modifier in `star.five.stroked.outlined`. Maybe we could use `.laced` for that? Now that we no longer use this modifier for a different purpose.
- `.three` variants and `.four` variants are the only ones for which the bare `.filled` variant is heavier than the `.filled.medium` variant.
  - I know you followed Unicode, but I think we should try to be consistent before trying to follow Unicode.
- Maybe U+2726 ✦ BLACK FOUR POINTED STAR could be added (under which name, though?).

T0mstone:
No worries!

> I still don't like the `.outlined` modifier in `star.five.stroked.outlined`. Maybe we could use `.laced` for that? Now that we no longer use this modifier for a different purpose.

I think I said this before, but do you have any evidence that it's actually laced other than just the font happened to design it that way? I'm all for this change if that concern could be dealt with.

> `.three` variants and `.four` variants are the only ones for which the bare `.filled` variant is heavier than the `.filled.medium` variant.
I know you followed Unicode, but I think we should try to be consistent before trying to follow Unicode.

Okay, sure, we already changed the weight names for the eight-pointed ones, so why not these as well.

> Maybe U+2726 ✦ BLACK FOUR POINTED STAR could be added

In that case we should also add U+2727 ✧ WHITE FOUR POINTED STAR and U+2BCC ⯌ LIGHT FOUR POINTED BLACK CUSP to U+2BCF ⯏ ROTATED WHITE FOUR POINTED CUSP.

src/modules/sym.txt
```diff
@@ -712,8 +712,53 @@ parallelogram
...
+  .five.filled.small ⭑
+  .five.filled.light 🟉
+  .five.filled.heavy 🟊
+  .five.filled.l ⯨
+  .five.filled.r ⯩
+  .five.filled.outlined ✭
+  .five.filled.outlined.heavy ✮
+  .five.pinwheel ✯
+  .five.crossing ⛤
+  .five.crossing.inv ⛧
```
MDLC01, line 731: I'm realizing this now: maybe this should be `.b` instead of `.inv` (as in, pointing to the bottom), and the other `.five` variants would have an additional `.t` modifier. I prefer it the way it is now, but I'm so lost with all those stars I figured it may be good to share this idea.

MDLC01:
> > `r.five.stroked.outlined`. Maybe we could use `.laced` for that? Now that we no longer use this modifier for a different purpose.
> 
> I think I said this before, but do you have any evidence that it's actually laced other than just the font happened to design it that way? I'm all for this change if that concern could be dealt with.

I am realizing now how annoying this is: the `.crossed.{l,r}` variants _are_ laced, and this one is too. With this realization in mind, I now dislike the `.l` and `.r` modifiers for `star.five.crossing`. Maybe we can just remove three variants for now and leave them for later?

> > Maybe U+2726 ✦ BLACK FOUR POINTED STAR could be added
> 
> In that case we should also add U+2727 ✧ WHITE FOUR POINTED STAR and U+2BCC ⯌ LIGHT FOUR POINTED BLACK CUSP to U+2BCF ⯏ ROTATED WHITE FOUR POINTED CUSP.

If you have good name ideas, let's add them. Otherwise, we can keep them for later.

T0mstone:
I just noticed that 🔯 is an emoji and seems to *only* be an emoji (it has no text presentation sequence), so I removed it.

PR #31, opened 2025-01-21
Add power symbols (Proposal 5)

T0mstone:
This PR implements Proposal 5 from the [Symbols Proposal Document](https://typst.app/project/pHN_iIPjeuCyI_N2JZB84o), although I tweaked it a little.

In particular, my tweaks are based on https://en.wikipedia.org/wiki/Power_symbol#Unicode,
adding `power.off` with value U+2B58 HEAVY CIRCLE and changing the plain `power` to `power.standby` (since that's what it represents).

IIUC, if the symbol doesn't have a character, it defaults to its first variant, right? That way, `power` without modifiers is also what it is in the proposal.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -540,6 +540,12 @@ join ⨝
   .l.r ⟗
 degree °
 smash ⨳
+power
+  .standby ⏻
+  .on ⏽
+  .off ⭘
+  .on.off ⏼
+  .sleep ⏾
 
 // Currency.
 bitcoin ₿


MDLC01:
I'm unsure those symbols should be added. And I'm even more unsure about `power.sleep`, which comes from a different set (IEEE 1621-2004, while the others are from ISO 7000:2012), and is much less widely used.

> This PR implements Proposal 5 (iteration 1) from the [Symbols Proposal Document](https://typst.app/project/pHN_iIPjeuCyI_N2JZB84o), although I tweaked it a little.

I edited the document to create a second iteration of the proposal, which should match this PR.

> adding `power.off` with value U+2B58 HEAVY CIRCLE and changing the plain `power` to `power.standby` (since that's what it represents).

Additionally, it U+2B58 has the informative alias "power off symbol."

> IIUC, if the symbol doesn't have a character, it defaults to its first variant, right?

Right.

mkorje (approved):
I don't really see any issue adding these symbols. Even though `power.sleep` comes from a different set, they were all added together to Unicode.

T0mstone:
I'll repeat here what I've said on Discord:
> It's true that I don't see a use for them in written documents, but they're really universally recognized symbols, so I'd say adding them doesn't hurt.

MDLC01 (approved):
As I expressed before, I don't really like those symbols, especially `power.sleep`. But given the slow rate of development here, I won't make you wait for someone else to approve this PR.

PR #33, opened 2025-01-21
Add root symbols (Proposal 13)

T0mstone:
This PR implements Proposal 13 (iteration 1) from the [Symbol Proposal Document](https://typst.app/project/pHN_iIPjeuCyI_N2JZB84o).

It's a draft since the document explicitly mentions being unsure of whether to add this, and I agree, but it easy to just make this change so we already have it available in case we do decide to accept it.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -254,6 +254,9 @@ times ×
   .square ⊠
   .triangle ⨻
 ratio ∶
+root √
+  .cube ∛
+  .fourth ∜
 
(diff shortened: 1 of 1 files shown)


MDLC01:
I don't think we should add those symbols for now.

T0mstone:
Yeah, let's let this sit. We can always come back to it later.

I've edited the PR description a bit. The last sentence had been really unclear in communicating my intentions.

PR #38, opened 2025-01-21
Add `to` and `mapsto` (Proposal 24)

T0mstone:
This PR implements Proposal 24 (iteration 2) from the [Symbol Proposals Document](https://typst.app/project/pHN_iIPjeuCyI_N2JZB84o).

I also added `to` as per our [Discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1331253538331951115).

This is a draft since it would be better done with the tools from #27.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -497,6 +497,9 @@ because ∵
 qed ∎
 
 // Function and category theory.
+to →
+mapsto ↦
+  .long ⟼
 compose ∘
 convolve ∗
 multimap ⊸
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -497,7 +497,6 @@ because ∵
 qed ∎
 
 // Function and category theory.
-to →
 mapsto ↦
   .long ⟼
 compose ∘


MDLC01:
Why not also adding `to.long`?

MDLC01:
> There's also still the option of not using aliases for these, but I feel like we should avoid duplication if possible.

I don't think the way we decide to implement a symbol (using aliases or not), or the way we implement our alias system, should dictate what we decide to add.

I don't know if we should add `to`, or `to.long`, but we should not take into consideration the way our alias system works when adding them. The final product should be good for the end users. Having good names is a higher priority than not duplicating some characters.

Just to be clear, I'm not saying that we should or shouldn't add `to` or `to.long`, just that this decision should be independent of the alias system.

> The name "to" also isn't nearly as semantic as "mapsto", so it doesn't really add much.

This is a good point.

T0mstone:
Given that you agreed to my last point, I've removed `to` again now.

MDLC01 (approved):

mkorje (approved):

PR #46, opened 2025-02-03
Codify and Resolve modifiers

### LINKED ISSUE #30: Resolve modifiers
As it stands now, the concept of a "modifier" doesn't really exist in codex itself. It only knows about _variants_. The closest thing is the syntax validation in `build.rs` (and the partially applied modifier alias code from #27).

Despite the decision from https://github.com/typst/codex/issues/5#issuecomment-2529659410 that codex shouldn't really aim to be independent for now, I think the code to resolve "symbol accessors" should be in codex instead of typst, so that the way modifiers work is formalized at our API boundary.

The way I envision this would be a `get(&str)` method on `Symbol`, so that you could resolve e.g. `lt.eq` by finding the `lt` symbol and calling `get("eq")` on it.

T0mstone:
Closes #30.

The most important idea of this PR is that it codifies what a "modifier" is.
In that aspect, it stands on its own and makes sense even without considering the context of usage in typst.

For this, there is a new `ModifierSet` type that contains all the logic for comparing and composing modifiers.
This new type is designed to be generic, so that typst can reuse it with `EcoString`.
(In particular, typst's `symbol::Repr::Modified` would change to `Modified(Arc<(List, ModifierSet<EcoString>)>)`)
Some private functions from `typst-library/src/foundations/symbol.rs` are lifted to public methods on `ModifierSet`:
- `parts` becomes `ModifierSet::iter`
- `contained` becomes `ModifierSet::contains`
- `find` becomes `ModifierSet::best_match_in` (with reversed argument order), which is generic since it was really easy to make it so and this prepares it for when we let symbols have more than one `char` per variant (cf. #25)

I've also taken the liberty of consolidating the types from `lib.rs` and `build.rs`. It bugged me that they were duplicated and this change didn't really do anything in terms of code size while arguably improving maintainability (unless you don't like macros, but even among macros this one is very tame).

There are some new functions (`get`, `variants` and `modifiers`) on `codex::Symbol` that somewhat mirror the ones on typst's `Symbol`, but  aren't intended to replace them right now (for why, see the next paragraph). They just made sense to add anyway.
The only changes made to typst would be to start using `ModifierSet`, which would basically have the effect of having moved the relevant code from typst to codex.

Actually integrating typst's `symbol::Repr::Modified` into codex proved too complicated, so I left that for later (if we even need it). Further considerations regarding the integrating of this into typst is left to a future PR at typst/typst (only if this one is accepted, of course).


MDLC01:
Like with #27, we can wait until Typst 0.13.0 is released to open the discussion and decide whether this should be part of Codex.

MDLC01:
Related: #51.

src/shared.rs
```diff
@@ -0,0 +1,148 @@
...
+        if !self.0.is_empty() {
+            self.0 += ".";
+        }
+        self.0 += m;
+    }
+
+    /// Iterate over the list of modifiers in an arbitrary order.
+    pub fn iter(&self) -> impl Iterator<Item = &str> {
+        self.0.split('.').filter(|s| !s.is_empty())
+    }
```
MDLC01, line 102: Why not implementing `IntoIterator` instead?

src/shared.rs
```diff
@@ -0,0 +1,148 @@
...
+            let mut total = 0;
+            for modifier in candidate.0.iter() {
+                if self.contains(modifier) {
+                    matching += 1;
+                }
+                total += 1;
+            }
+
+            let score = (matching, core::cmp::Reverse(total));
+            if best_score.map_or(true, |b| score > b) {
```
MDLC01, line 140:
```suggestion
            if best_score.is_none_or(|b| score > b) {
```

CI uses stable Rust so I think this should be fine.

src/shared.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /// Construct a modifier set from a string,
+    /// where modifiers are separated by the character `.`.
+    ///
+    /// It is not unsafe to use this function wrongly, but it can produce
+    /// unexpected results down the line. Correct usage should ensure that
+    /// `s` does not contain any empty modifiers (i.e. the sequence `..`)
+    /// and that no modifier occurs twice.
+    pub fn new_unchecked(s: S) -> Self {
```
MDLC01, line 67: For use in Typst, I think we could benefit from having a checked `new` constructor. Although I guess Codex shouldn't have to worry about what a valid identifier is (e.g., a valid identifier should not contain a space), so Typst will have to check the strings either way.

src/lib.rs
```diff
@@ -1,10 +1,23 @@
...
 Human-friendly notation for Unicode symbols.
+
+## Model
+A [`Symbol`] is a collection of one or more _variants_.
+Each variant is identified by a set of _modifiers_ (see [`ModifierSet`])
+and has a single character as its value.
+The modifiers themselves can in principle be any non-empty strings
+that don't contain the character `.`, but codex only defines
+ones that are entirely made of ASCII alphabetical characters.
 */
```
MDLC01, line 11:
```suggestion
//! Human-friendly notation for Unicode symbols.
//! 
//! ## Model
//! A [`Symbol`] is a collection of one or more _variants_.
//! Each variant is identified by a set of _modifiers_ (see [`ModifierSet`])
//! and has a single character as its value.
//! The modifiers themselves can in principle be any non-empty strings
//! that don't contain the character `.`, but codex only defines
//! ones that are entirely made of ASCII alphabetical characters.
```

Based on https://github.com/typst/codex/pull/20#discussion_r1966738841.

src/shared.rs
```diff
@@ -0,0 +1,148 @@
...
+        Self(s)
+    }
+
+    /// Construct an empty modifier set.
+    pub fn empty() -> Self
+    where
+        S: Default,
+    {
+        Self(S::default())
+    }
```
MDLC01, line 77: Why not implementing `Default` when `S: Default` instead? Technically, some string types could default to non-empty strings.

src/lib.rs
```diff
@@ -21,38 +34,45 @@ impl Module {
...
-pub struct Binding {
-    /// The bound definition.
-    pub def: Def,
-    /// A deprecation message for the definition, if it is deprecated.
-    pub deprecation: Option<&'static str>,
+impl<'a> ModifierSet<&'a str> {
+    /// Iterate over the list of modifiers with the original lifetime.
+    pub fn to_iter(self) -> impl Iterator<Item = &'a str> {
+        self.0.split('.').filter(|s| !s.is_empty())
+    }
```
MDLC01, line 41: Why is this not in `shared.rs`?

src/lib.rs
```diff
@@ -1,10 +1,23 @@
 /*!
 Human-friendly notation for Unicode symbols.
+
+## Model
+A [`Symbol`] is a collection of one or more _variants_.
+Each variant is identified by a set of _modifiers_ (see [`ModifierSet`])
```
MDLC01, line 6:
```suggestion
Each variant is identified by a set of [_modifiers_ ](ModifierSet)
```

MDLC01 (commented):
I left a lot of comments and suggestions. Of course, feel free to disagree.

I really like the `ModifierSet` abstraction. If feels very clean. I'm not a fan of the `include!` trick and the macro, but it's probably fine.

Regarding the `ModifierSet::iter` and `ModifierSet::<&'a str>::to_iter` methods, I feel like this could be achieved more idiomatically with two separate `impl<'a, S: Deref<Target = str>> IntoIterator for &'a ModifierSet<S>` and `impl<'a> IntoIterator for ModifierSet<&'a str>`.

src/shared.rs
```diff
@@ -0,0 +1,148 @@
+use std::ops::{AddAssign, Deref};
+
+macro_rules! declare_types {
+    ($(<$lt:lifetime>)?
+        $(derive($($Der:ident),*),)?
+        str = $s:ty,
+        List = $List:ident<_>
```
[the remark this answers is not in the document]
T0mstone: Your suggestion wouldn't allow substituting a concrete type for `$T` (except for maybe defining a type alias `type $T = ...`, but that's just as hacky as the `StaticSlice` one)

src/shared.rs
T0mstone: That would require spelling out the type instead of just using `impl Iterator`.
If only we had TAIT already...🥲

src/shared.rs
T0mstone: Yep, I wrote this before the recent MSRV bump.

src/shared.rs
T0mstone: Yep, that's exactly why I didn't write a checked one.

src/lib.rs
T0mstone: It's a relief rust-analyzer can do this in one click😅

src/shared.rs
T0mstone: Hmmm, in that case the default value could also be invalid...

I guess implementing `Default` is just more idiomatic either way. I'll write a doc-comment of caution; That'll probably be enough.

src/lib.rs
T0mstone: uhh, probably just because the build-script doesn't need it and because it's a concrete impl and not a generic one. I'll just move it, no big deal.

src/shared.rs
MDLC01: Codex CI is separate, and used latest stable Rust as of now.

T0mstone: ah, the more you know

MDLC01:
I feel like this would be more idiomatic, but it doesn't matter that much.

Also, it may be useful for later to have something like `impl<S: for<'a> std::ops::AddAssign<&'a str>> FromIterator for ModifierSet<S>`. But if there is any issue implementing that, it does not matter for now.

src/shared.rs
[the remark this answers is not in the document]
MDLC01: Ok that makes sense.

T0mstone:
> Also, it may be useful for later to have something like `impl<S: for<'a> std::ops::AddAssign<&'a str>> FromIterator for ModifierSet<S>`.

For that, there's the issue of an empty iterator. Should it just require `S: Default` and assume that the default is the empty string?
Apart from that question, the implementation is pretty easy.

MDLC01 (approved):
Thanks!

MDLC01:
@laurmaedje you may want to review this PR

MDLC01:
@laurmaedje I'm taking the liberty to ping you again as you are more active in PR reviews now.

laurmaedje:
Yes, I'll try to get to this in the next few days. (If I still haven't done it in a week or so, please feel free to ping be again and then I'll do it right away.)

laurmaedje:
> I've also taken the liberty of consolidating the types from lib.rs and build.rs. It bugged me that they were duplicated and this change didn't really do anything in terms of code size while arguably improving maintainability (unless you don't like macros, but even among macros this one is very tame).

I'm not a fan of macros, so I'm not sure I like this change, but in any case, I think it would best to discuss this separately from this PR. The fact that I was unsure whether I liked that is part of why it took me so long to respond here. If I see things correctly, the `ModifierSet` is also not needed in the build script, it's just there to make the type a bit more semantic, right?

T0mstone:
> If I see things correctly, the `ModifierSet` is also not needed in the build script, it's just there to make the type a bit more semantic, right?

Not yet, but I'd need it for #27...

> I think it would be best to discuss this separately from this PR

I'll see what I can do, but I'm not happy about this.

T0mstone:
I've removed the macro, but kept `ModifierSet` shared since a `ModifierSet` type is needed in `build.rs` for debug-printing anyway.

Having done this change directly, I'm now actually unsure myself whether the macro was worth it, in a "both sides are bad" kind of way, so I probably won't make another PR for that, at least for now.

T0mstone:
Also, minor stylistic question, but should `shared.rs` be called `modifier_set.rs` now that it only contains one thing?

src/shared.rs
```diff
@@ -0,0 +1,126 @@
+use std::ops::Deref;
+
+/// A set of modifiers.
+#[derive(Debug, Copy, Clone)]
+// note: the visibility needs to be `pub(crate)`,
+// since build.rs outputs `ModifierSet(...)`
```
laurmaedje, line 6: I think this is outdated since it uses `new_unchecked`

laurmaedje (commented):
I think keeping it as `shared.rs` is fine and more future proof. Assuming all the Typst functionality can be implemented with this, the API looks fine to me!

T0mstone:
> Assuming all the Typst functionality can be implemented with this, the API looks fine to me!

Cool, I'll get working on a PR to typst then.

src/shared.rs
T0mstone: No, the point where it gets output is https://github.com/typst/codex/blob/34edb0933b20458dbe8b97a60838697be533cc89/build.rs#L217
which uses the debug-formatting, which outputs `ModifierSet(<string>)`.
Changing that would of course be possible, but it'd require constructing the slice manually instead of just using `{:?}` there.

T0mstone:
`typst/docs` needs a way to `Display` a `ModifierSet`, so I think the most idiomatic way of achieving that is just bringing `as_str` back.

T0mstone:
Okay, bug fixed. This issue has brought up the need for some unit tests for `ModifierSet`. I'll write some more of those now, but feel free to also suggest some test cases.

T0mstone:
Should be ready to merge now.

MDLC01:
@laurmaedje all your concerns were addressed

Enivex:
> @laurmaedje all your concerns were addressed

I think he's been busy with other things in recent weeks.

laurmaedje:
Indeed. I will definitely return to this and all other PRs in the coming weeks.

src/shared.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl<S: Deref<Target = str>> ModifierSet<S> {
+    /// Constructs a modifier set from a string, where modifiers are separated by
+    /// the character `.`.
+    ///
+    /// It is not unsafe to use this function wrongly, but it can produce
+    /// unexpected results down the line. Correct usage should ensure that `s`
+    /// does not contain any empty modifiers (i.e. the sequence `..`) and that
+    /// no modifier occurs twice.
+    pub fn new_unchecked(s: S) -> Self {
```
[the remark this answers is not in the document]
T0mstone: > Also new typically creates something empty in Rust lingo.

Not necessarily, e.g. [`Path::new`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.new).

T0mstone:
My rationale for choosing this naming for the contended functions: `_raw` should still convey the feeling that you have to do the checks yourself before calling the function while hopefully not sounding as `unsafe`y.
Note that for `insert_raw`, the argument should _not_ contain any `.`, hence it doesn't have `_dotted` in the name.

Also please review the reworded docs for these two functions to make sure I didn't accidentally make them worse in your eyes.

T0mstone:
I also took the liberty of running `cargo clippy --fix`. I hope that's not an issue; The yellow underlines were bothering me.😅

laurmaedje (approved):

PR #50, opened 2025-02-19
Add struck arrows

MDLC01:
This PR adds struck arrows, as well as twohead tailed arrows.

For singly struck arrows, I chose the `.struck` modifier, which is consistent with its use in `parallel.struck` in contrast to `parallel.not`. Unicode also defines doubly struck arrows. For those, an additional modifier needs to be created, and i am unsure of the name to use. Maybe `.doublestruck`, or `.dstruck`? I marked the PR as a draft for this reason.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -752,6 +752,7 @@ arrow
   .r.double.long ⟹
   .r.double.long.bar ⟾
   .r.double.not ⇏
+  .r.double.struck ⤃
   .r.filled ➡
   .r.hook ↪
   .r.long ⟶
@@ -762,12 +763,16 @@ arrow
   .r.squiggly ⇝
   .r.stop ⇥
   .r.stroked ⇨
+  .r.struck ⇸
   .r.tail ↣
+  .r.tail.struck ⤔
   .r.tilde ⥲
   .r.triple ⇛
   .r.twohead ↠
   .r.twohead.bar ⤅
+  .r.twohead.struck ⤀
   .r.twohead.tail ⤖
+  .r.twohead.tail.struck ⤗
   .r.wave ↝
   .l ←
   .l.bar ↤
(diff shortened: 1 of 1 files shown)


Enivex:
`struck.double` or `struck.twice`?

T0mstone:
> `struck.double` or `struck.twice`?

`arrow.double` is already something else. This is a case where our commutative modifiers are a limitation, since the "double" would be a modifier that applies to another modifier.

Enivex:
Are modifiers commutative though?

T0mstone:
> Are modifiers commutative though?

Yesn't.
They are in typst, that's how `circle.filled.big` is the same as `circle.big.filled`.
For codex by itself, modifiers don't exist. We only have variants until/unless #46 is merged.

knuesel:
Maybe `.struck-twice` ?

MDLC01:
The dash would prevent it from being easily usable in math

knuesel:
Ah right, I confusedly imagined this would be OK after the change to math argument parsing in Typst 0.13 but obviously not!

In favor of `dstruck`: we could consistently use a `d` prefix for double modifiers, like we use `n` for negation in `.neq`, `.neqiv`, `.ndilde`.

Or we could aim for consistency with `arrow.r.twohead` and use `arrow.r.twostroke` (as in the Unicode name 	Rightwards Arrow with Double Vertical Stroke).  Then it would make sense to rename `arrow.r.struck` to `arrow.r.stroke`...

MDLC01:
`arrow.r.stroke` is too similar to `arrow.r.stroked` imo.

MDLC01:
I went with `.dstruck`. As mentioned above, it creates a consistency with `.neq`. If anything, `.twohead` is the outlier here, although I don't think it would be necessary to change it.

T0mstone (approved):

mkorje (approved):

PR #54, opened 2025-03-11
Add names for open-headed arrows

zacharie-moughanim:
Add names for open-headed arrows

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -783,6 +783,7 @@ arrow
   .r.twohead.tail ⤖
   .r.twohead.tail.struck ⤗
   .r.twohead.tail.dstruck ⤘
+  .r.open ⇾
   .r.wave ↝
   .l ←
   .l.bar ↤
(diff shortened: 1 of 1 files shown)


mkorje (approved):

T0mstone:
How about using `.stroked` instead of `.open`?

T0mstone:
Ah, no, that's already something else, oops. Still, I'd like it for there to be some sort of connection to `stroked` instead of just introducing an entirely unrelated modifier.

MDLC01:
Could be `.strokedhead` (there is already `.twohead`). But it's a longer name.

mkorje:
Yeah that's a bit of a mouthful imo. If we want to keep it related to `stroked`, how about something like `.stroked.head`?

MDLC01:
`.stroked.head` sounds weird to me. In particular, note that `.head` alone would fallback to that.

mkorje:
Ah I forgot about that, not so great then.

Enivex:
We already have a billion different meanings for stroked. Open is as good a name as any, and is what the official unicode name uses.

MDLC01 (approved):
Let's go with `.open` then

PR #56, opened 2025-04-12
Deprecate `shin`

mkorje:


MDLC01:
As a side note, there is another Hebrew letter that was used in maths but does not have a corresponding LTR symbol: ת (Tav).[^1] However, I do not think it is used in modern papers outside of this specific work by Cantor.

[^1]: https://en.wikipedia.org/wiki/Tav_(number)

Enivex:
> As a side note, there is another Hebrew letter that was used in maths but does not have a corresponding LTR symbol: ת (Tav).[1](#user-content-fn-1-49c01f8720ac286de05e560b0e090d2a) However, I do not think it is used in modern papers outside of this specific work by Cantor.
> ## Footnotes
> 
>     1. https://en.wikipedia.org/wiki/Tav_(number) [↩](#user-content-fnref-1-49c01f8720ac286de05e560b0e090d2a)

That's not in codex currently no?

MDLC01:
Indeed it isn't

Enivex (changes requested):
We may want to explain the reason for deprecation?

mkorje:
Would something like "`shin` is deprecated and will be removed as it does not have a corresponding LTR symbol" work? It's maybe a bit too verbose.

MDLC01:
> It's maybe a bit too verbose.

And most importantly, most users would have no idea what this means.

Enivex (approved):

Enivex:
I guess it's fine

dccsillag:
It might be worth it to point the users to sha (since that's what shin would mainly be used for)? (We do have sha in codex, right? I tried to check now but couldn't find it...)

Enivex:
> It might be worth it to point the users to sha in the deprecation note (since that's what shin would mainly be used for)? (We do have sha in codex, right? I tried to check now but couldn't find it...)

No, we do not currently have sha

MDLC01:
Maybe it should be added in this PR then? We can keep other Cyrillic letters for later, but at least if we can provide a sort of replacement it would be great.

MDLC01:
I had not seen https://github.com/typst/codex/pull/67 when writing the previous comment. Then let's just mention `sha` in the deprecation message here.

MDLC01:
Either are fine I think

dccsillag:
@emilyyyylime 's suggestion sounds slightly better to me

mkorje:
I've changed it to @emilyyyylime 's suggestion,

Enivex:
Reminder: It's breaking, so don't merge without feedback from boss man

PR #58, opened 2025-04-16
Add U+00A2 `¢` and U+20B5 `₵`

WagyuDEV:
The `¢` "cent" symbol is used when referring to fractional dollars in US currency. The `₵` "cedi" (Ghanaian Cedi) is the legal tender of the Republic of Ghana.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -573,9 +573,9 @@ power
 
 // Currency.
 bitcoin ₿
-dollar $
 cedi ₵
 cent ¢
+dollar $
 euro €
 franc ₣
 lira ₺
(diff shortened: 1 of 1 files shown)


mkorje (approved):

src/modules/sym.txt
```diff
@@ -574,6 +574,8 @@ power
 // Currency.
 bitcoin ₿
 dollar $
+cedi ₵
+cent ¢
```
T0mstone, line 578: Small nitpick: These should be above `dollar` to keep the currency symbols in alphabetical order.

MDLC01 (approved):

PR #59, opened 2025-04-26
Add remaining currency symbols, deprecate franc

Enivex:
Adds the following currency symbols:

Armenian Dram Sign - dram (Currency of Armenia)
Afghani Sign - afghani (Currency of Aghanistan)
Bengali Rupee Sign - taka (Currency of Bangladesh)
Tamil Rupee Sign - rupee.tamil (Alternative symbol for Indian rupee in the Tamil alphabet)
Thai Currency Symbol Baht - baht (Currency of Thailand)
Khmer Currency Symbol Riel - riel (Currency of Cambodia)
Colon Sign - colon.currency (Currency of Costa Rica)
Naira Sign - naira (Currency of Nigeria)
Rupee sign - rupee.general (Currency of Mauritius, Nepal, Pakistan, Seychelles, Sri Lanka)
New Sheqel Sign - shekel (Currency of Israel)
Dong Sign - dong (Currency of Vietnam)
Kip Sign - kip (Currency of Laos)
Tugrik sign - tugrik (Currency of Mongolia)
Guarani Sign - guarani (Currency of Paraguay)
Hryvnia Sign - hryvnia (Currency of Ukraine)
Tenge Sign - tenge (Currency of Kazakhstan)
Manat Sign - manat (Currency of Azerbaijan)
Lari Sign - lari (Currency of Georgia)
Wancho Ngun Sign - rupee.wancho (Alternative symbol for Indian rupee in the Wancho alphabet)
Nko Dorome Sign - dorome (Currency symbol in the N'ko alphabet)
Nko Taman Sign - taman (Currency symbol in the N'ko alphabet)

Deprecates the following symbol:
French Franc Sign - franc (Not only is the currency no longer in use, the symbol was only a proposal and never used)

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -112,6 +112,7 @@ backslash \
   .not ⧷
 co ℅
 colon :
+  .currency ₡
   .double ∷
   .tri ⁝
   .tri.op ⫶
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -574,23 +574,39 @@ smile ⌣
 frown ⌢
 
 // Currency.
+afghani ؋
+baht ฿
 bitcoin ₿
 cedi ₵
 cent ¢
 dollar $
+dong ₫
+dorome ߾
+dram ֏
 euro €
 @deprecated: `franc` is deprecated, unadopted symbol for currency no longer in use
 franc ₣
+guarani ₲
+hryvnia ₴
+kip ₭
+lari ₾
 lira ₺
+manat ₼
+naira ₦
 peso ₱
 pound £
+riel ៛
 ruble ₽
 rupee
   .indian ₹
   .general ₨
   .tamil ௹
   .wancho 𞋿
+shekel ₪
 taka ৳
+taman ߿
+tenge ₸
+tugrik ₮
 won ₩
 yen ¥
 
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -579,6 +579,7 @@ cedi ₵
 cent ¢
 dollar $
 euro €
+@deprecated: `franc` is deprecated, unadopted symbol for currency no longer in use
 franc ₣
 lira ₺
 peso ₱
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -585,7 +585,12 @@ lira ₺
 peso ₱
 pound £
 ruble ₽
-rupee ₹
+rupee
+  .indian ₹
+  .general ₨
+  .tamil ௹
+  .wancho 𞋿
+taka ৳
 won ₩
 yen ¥
 


Enivex:
I purposely avoided adding symbol names for currencies no longer in active use.

T0mstone:
Thoughts on also adding U+00A4 ¤ CURRENCY SIGN as `currency`? Or maybe as `scarab` (as [Wikipedia says](https://en.wikipedia.org/wiki/Currency_sign_(generic)) it's also called)?

Enivex:
> Thoughts on also adding U+00A4 ¤ CURRENCY SIGN as `currency`? Or maybe as `scarab` (as [Wikipedia says](https://en.wikipedia.org/wiki/Currency_sign_(generic)) it's also called)?

I did consider that, but it seems to be a legacy symbol. Is it actually used?

src/modules/sym.txt
```diff
@@ -112,6 +112,7 @@ backslash \
   .not ⧷
 co ℅
 colon :
+  .currency ₡
```
mkorje, line 115: This confused me a bit at first and I'm not sure this is a good name for the symbol, but I understand why you've done it.

An idea: maybe we could put all the currency symbols as a modifier under `currency`?

MDLC01: That would make the names much longer, for no benefit in most cases. I don't want to have to write, e.g., `currency.euro`. `euro` should be enough.

mkorje: I figure though in most cases you would have the currency symbol you use most on your keyboard anyways, so the long name isn't too terrible as you wouldn't use it too much. But yeah it's not ideal.

Enivex: > I figure though in most cases you would have the currency symbol you use most on your keyboard anyways, so the long name isn't too terrible as you wouldn't use it too much. But yeah it's not ideal.

I think that many of these symbols aren't on keyboard layouts

Enivex: > That would make the names much longer, for no benefit in most cases. I don't want to have to write, e.g., `currency.euro`. `euro` should be enough.

It would be even worse than that, since you most likely would want the text font. So `#sym.currency.euro`

emilyyyylime:
What about putting currencies under a module so that you can import what you need?

Enivex:
> What about putting currencies under a module so that you can import what you need?

I don't think there are any plans to have more categories than general symbols and emojis. In any case I think that would be a discussion not really specific to this PR?

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+manat ₼
+naira ₦
 peso ₱
 pound £
+riel ៛
 ruble ₽
-rupee ₹
+rupee
+  .indian ₹
+  .general ₨
```
MDLC01, line 603: Not sure `.general` is a good name here. Not sure this should even be added either. If it is just a precomposed ligature of R and s, why not just use "Rs"?

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+  .tamil ௹
+  .wancho 𞋿
```
MDLC01, line 605:
```suggestion
wancho 𞋿
```

Why not just `wancho`?

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+shekel ₪
+taka ৳
+taman ߿
```
MDLC01, line 608: Do you have sources that this is still in use?

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
 
 // Currency.
+afghani ؋
+baht ฿
 bitcoin ₿
 cedi ₵
 cent ¢
 dollar $
+dong ₫
+dorome ߾
```
MDLC01, line 585: Do you have sources that this is still in use?

src/modules/sym.txt
Enivex: I took general from https://en.wikipedia.org/wiki/Rupee_sign?wprov=sfla1. Not sure what else to call it. We could call it just rupee, but that would require a breaking change. Plus the Indian rupee has more users, even if this one is used in more countries.

As for why the precomposed symbol: Unlike other such symbols where usage seems to be discouraged, I couldn't find anything for this symbol. It seems that it's actually used in practice. It's also a semantic thing for screen readers and such.

src/modules/sym.txt
Enivex: Wancho is the name of a language, not the currency. If we wanted to use a different name, then `ngun` would probably be the way to go (though I couldn't find a straight answer to its meaning). However, it does seem to just be an alternative symbol for the Indian rupee in the wancho alphabet, which is why I made the choice I did.

src/modules/sym.txt
Enivex: It's a modern alphabet, and the Unicode proposal doesn't make any mention of it not being in use.

src/modules/sym.txt
Enivex: See the answer to taman

src/modules/sym.txt
mkorje: The core spec has this to say about U+20A8 ₨ RUPEE SIGN
> ... It is distinguished from U+20A8 RUPEE SIGN, which is an older symbol not formally tied to any particular currency
Also listed with the description
> India, unofficial legacy practice
in the code chart.

mkorje: Also, on the Wikipedia page where it lists countries where the Rs sign is used, actually going to those currencies pages shows either a different symbol or that the symbol is Re/Rs, not this character?

mkorje:
Reading the [Unicode core spec](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G26389), there's also the Som Sign U+20C0 ⃀, the official currency symbol of the Kyrgyz Republic.

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+@deprecated: `franc` is deprecated, unadopted symbol for currency no longer in use
 franc ₣
+guarani ₲
+hryvnia ₴
+kip ₭
+lari ₾
 lira ₺
+manat ₼
+naira ₦
 peso ₱
```
mkorje, line 597: The core spec has this to say:
> Dollar and Peso. The dollar sign (U+0024) is used for many currencies in Latin America and elsewhere. In particular, this use includes current and discontinued Latin American peso currencies, such as the Mexican, Chilean, Colombian and Dominican pesos. However, the Philippine peso uses a different symbol found at U+20B1.

So maybe peso should point to $ and peso.philippine point to ₱.

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+  .general ₨
+  .tamil ௹
+  .wancho 𞋿
+shekel ₪
+taka ৳
+taman ߿
+tenge ₸
+tugrik ₮
 won ₩
 yen ¥
```
mkorje, line 612: Slightly unrelated to the rest of the PR, but yuan should probably be added as an alias for ¥.
> Yen and Yuan. Like the dollar sign and the pound sign, U+00A5 YEN SIGN has been used as the currency sign for more than one currency. The double-crossbar glyph is the official form for both the yen currency of Japan (JPY) and for the yuan (renminbi) currency of China (CNY). This is the case, despite the fact that some glyph standards historically specified a single-crossbar form, notably the OCR-A standard ISO 1073-1:1976, which influenced the representative glyph in various character set standards from China. In the Unicode Standard, U+00A5 YEN SIGN is intended to be the character for the currency sign for both the yen and the yuan, independent of the details of glyphic presentation.

src/modules/sym.txt
Enivex: > The core spec has this to say about U+20A8 ₨ RUPEE SIGN
> > ... It is distinguished from U+20A8 RUPEE SIGN, which is an older symbol not formally tied to any particular currency
> 
> Also listed with the description
> > India, unofficial legacy practice
> 
> in the code chart.

Older here is in relation to the Indian rupee symbol, which is (unsurprisingly) not in use outside India

Enivex: > Also, on the Wikipedia page where it lists countries where the Rs sign is used, actually going to those currencies pages shows either a different symbol or that the symbol is Re/Rs, not this character?

It's not universally used, no. However I've not seen any indication that's it's not actually in use.

Enivex:
> Reading the [Unicode core spec](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G26389), there's also the Som Sign U+20C0 ⃀, the official currency symbol of the Kyrgyz Republic.

Good spot. I must've looked at outdated lists. I'll add it.

src/modules/sym.txt
Enivex: It's the only peso (as far as I know) with its own symbol.

There's several other currencies using the dollar symbol that would then presumably also need to be added. I don't think this is a good idea.

The goal isn't to add every currency, but every currency _symbol_.

src/modules/sym.txt
Enivex: See my comment on dollar vs peso

src/modules/sym.txt
mkorje: I clearly didn't look very hard as the Sri Lankan rupee has it as the symbol.

src/modules/sym.txt
mkorje: I'm not sure how accurate this is, but both taman and dorome are listed as part of the [historical currency symbols on wikipedia](https://en.wikipedia.org/wiki/Currency_symbol#List_of_historical_currency_symbols)

src/modules/sym.txt
```diff
@@ -574,17 +575,39 @@ smile ⌣
...
+rupee
+  .indian ₹
+  .general ₨
+  .tamil ௹
+  .wancho 𞋿
+shekel ₪
+taka ৳
+taman ߿
+tenge ₸
+tugrik ₮
```
mkorje, line 610: Whilst Unicode uses the name Tugrik sign, it appears the name togrog is instead more common. [Wiktionary](https://en.wiktionary.org/wiki/tugrik#Usage_notes) has this to say
> Tugrik is the [transliteration](https://en.wiktionary.org/wiki/transliteration) used in the [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) standard for currency names, and is therefore widely used in business and banking. CIA World Factbook also recognizes togrog, and Wikipedia has chosen tögrög as the main entry.

mkorje: Togrog is also what the [Central Bank of Mongolia](https://www.mongolbank.mn/en/p/1392) uses.

src/modules/sym.txt
mkorje: Fair point

src/modules/sym.txt
Enivex: It's tricky, because ultimately they're all just different transliterations of the Mongolian name. The closest we can get to an "official" one is the ISO one.

src/modules/sym.txt
Enivex: > There appears to be only one other currency (other than peso and dollar) that uses the dollar symbol: the Macanese pataca (https://en.wikipedia.org/wiki/Currency_symbol#List_of_currency_symbols_currently_in_use).

Ok fair enough, if we're talking currencies with truly different names, then there are only three.

The question is whether we really want to start adding multiple names for the same currency symbols?

src/modules/sym.txt
mkorje: Hmm, ISO 4217 only assigns alphabetic and numeric currency codes, not the specific transliteration. I do also think that if the central bank uses the transliteration togrog that's a good indicator of what a user would use.

src/modules/sym.txt
mkorje: I don't think it's a terrible idea if we do start add multiple names. Especially when this situation with dollar appears to be the only one (and the yen and yuan).,]

Enivex:
I added the Kyrgysz currency and changed the name of the Mongolian currency.

However, I think @laurmaedje would have to make the decision of whether we should add the additional aliases "yuan" (to yen), "peso" (and rename "peso" to "peso.phillipine") and "pataca" (to dollar).

mkorje (approved):

Enivex:
I added the currency symbol after all, and renamed rupee.general to rupee.generic.

mkorje (approved):
I don't have an opinion on rupee.general or rupee.generic, either seems fine.

Issue #34, opened 2025-01-21
`circled` instead of `circle` (Proposal 15)


T0mstone:
This is the issue for Proposal 15 from the [Symbol Proposal Document](https://typst.app/project/pHN_iIPjeuCyI_N2JZB84o).

The text of the proposal (iteration 1) is reproduced here:

## Context

For circled symbols, we currently use the `.circle` modifier. This creates a name conflict in the case of #char("⨢") and #char(sym.plus.circle), which both ought to be named `plus.circle` under this convention.

## Proposal

Rename `.circle` modifiers that correspond to circled variants to `.circled`. Similarly, rename `.triangle` and `.square` to `.triangled` and `.squared` for consistency.

## Breaking changes

Many variants would become invalid. The exact list of such variants would be determined when implementing the proposal.

MDLC01:
I think adding a single letter is fine given this allows for consistency, and thus greater predictability of symbol names.

Also, note that other existing modifiers tend to use similar grammar, such as `.filled`, `.stroked`, or `.struck`.

A possible alternative is to keep the current names, and when there is an ambiguity, such as with ⊕ and ⨢, chose one to be the "default", and use an appropriate modifier to select the other one. For example, ⊕ would be `plus.circle`, and ⨢ would be `plus.circle.above`. In this case, this only works because `.above` does not already mean something for `plus`.

PR #60, opened 2025-05-16
Add `slash.circle`

arvidfm:
This adds the name `slash.circle` for the Unicode symbol ⊘, which was inexplicably missing despite seemingly every other circled operator being included. This symbol is sometimes used for Hadamard (element-wise) division, and is available as `\oslash` in LaTeX.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -168,6 +168,7 @@ section §
 semi ;
   .rev ⁏
 slash /
+  .circle ⊘
   .double ⫽
   .triple ⫻
   .big ⧸


MDLC01:
This is probably fine to add if you need it, but otherwise I would suggest waiting for a resolution to https://github.com/typst/codex/issues/34 before adding any more "symbols inside shapes". A (maybe exhaustive?) list of those symbols is provided in Section 3.4. of the [symbol Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT).

arvidfm:
I can't imagine that adding the symbol would make any breakage resulting from #34 substantially worse to deal with, considering any document that makes use of Hadamard division is almost certain to make even more frequent use of the Hadamard product, which is also affected. Having to wait for #34 to be resolved to have a nice way to type the Hadamard division operator seems a bit silly to me.

MDLC01:
Yeah that's what I meant by "It's probably fine to add if you need it": let's not block adding one symbol on some hypothetical future change if there is an actual need for said symbol. 👍

Enivex:
> I can't imagine that adding the symbol would make any breakage resulting from #34 substantially worse to deal with, considering any document that makes use of Hadamard division is almost certain to make even more frequent use of the Hadamard product, which is also affected. Having to wait for #34 to be resolved to have a nice way to type the Hadamard division operator seems a bit silly to me.

You would have to wait for the next version of typst regardless, and its likely that #34 would be resolved by then

arvidfm:
Oh, if it'll have been dealt with before the next typst release anyway, I don't personally care either way. I just assumed that #34 might be a while since it's been open for a few months already with seemingly no/little discussion.

Enivex:
Most of the discussion takes place on discord, and there's not a lot of incentive to move fast when there's no imminent release.

MDLC01:
We can probably keep this PR open and merge it before the next release if it becomes apparent that #34 will not be solved by then.

Enivex:
This is in conflict with the resolution picked in https://github.com/typst/codex/pull/62 by the way. Though that PR does not add additional symbols.

MDLC01 (approved):
Let's just merge this and @Enivex can update #62. This is probably easier and will reduce the open PR count.

PR #61, opened 2025-05-17
Deprecate alternate spellings of hebrew characters

Enivex:
We had two alternate spellings largely because there was no mechanism for deprecation at the time. I'm not opposed to maintaining more than one name for a symbol if they're fundamentally different and there's a good reason for it, but different romanizations is many steps too far in my opinion. A similar example came up recently for the Mongolian currency, which you can find references to as tugrik, tughrik, tugrug, togrog, tögrög, and I'm sure even more.

1. The spellings alef, bet, gimmel and dalet are, as far as I can tell, closer to modern hebrew romanization (actually, "gimmel" seems to be very obscure regardless). However, aleph, beth, gimel and daleth are almost universal in an academic setting, which is where these symbols are intended for.

A search for "alef" on arxiv brings up mostly a bunch of results relating to an author with the last name "Alef" https://arxiv.org/search/?query=alef&searchtype=all, while "aleph" has many relevant results https://arxiv.org/search/?query=aleph&searchtype=all (even excluding the matches to "\aleph").

This is similar to the situation with the greek letters, where "alfa" would have been preferred if we want by romanization.

2. LaTeX uses \aleph, \beth, \gimel and \daleth

3.  HTML uses &aleph, &beth, &gimel and &daleth

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1051,12 +1051,16 @@ Zeta Ζ
 // from Letterlike Symbols.
 // See https://github.com/typst/typst/pull/3375.
 aleph א
+@deprecated: `alef` is deprecated, use `aleph` instead
 alef א
 beth ב
+@deprecated: `bet` is deprecated, use `beth` instead
 bet ב
+@deprecated: `gimmel` is deprecated, use `gimel` instead
 gimmel ג
 gimel ג
 daleth ד
+@deprecated: `dalet` is deprecated, use `daleth` instead
 dalet ד
 shin ש
 
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1056,9 +1056,9 @@ alef א
 beth ב
 @deprecated: `bet` is deprecated, use `beth` instead
 bet ב
-@deprecated: `gimmel` is deprecated, use `gimel` instead
-gimmel ג
 gimel ג
+@deprecated: `gimmel` is deprecated, use `gimel` instead
+ gimmel ג
 daleth ד
 @deprecated: `dalet` is deprecated, use `daleth` instead
 dalet ד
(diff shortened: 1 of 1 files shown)


MDLC01:
> `alef` also conflicts with the Arabic letter, which we could add in the future along with the rest of the Arabic alphabet (after #20 is merged).

Ok this convinces me

MDLC01 (approved):

emilyyyylime:
I think this makes sense. Even in Israel in an academic setting you'd find 'aleph' more commonly than 'alef' I think.<sup>[<ins>_citation needed_</ins>]</sup>

src/modules/sym.txt
```diff
@@ -1051,12 +1051,16 @@ Zeta Ζ
...
 // See https://github.com/typst/typst/pull/3375.
 aleph א
+@deprecated: `alef` is deprecated, use `aleph` instead
 alef א
 beth ב
+@deprecated: `bet` is deprecated, use `beth` instead
 bet ב
+@deprecated: `gimmel` is deprecated, use `gimel` instead
 gimmel ג
 gimel ג
```
emilyyyylime, line 1061:
```suggestion
gimel ג
@deprecated: `gimmel` is deprecated, use `gimel` instead
 gimmel ג
 ```
Maybe change the order here so it's consistent with the other letters?

Comment (Enivex) [reply src/modules/sym.txt:1061]: Alright with me.

File: src/modules/sym.txt [hunk 2]
```diff
@@ -1051,12 +1051,16 @@ Zeta Ζ
...
 aleph א
+@deprecated: `alef` is deprecated, use `aleph` instead
 alef א
 beth ב
+@deprecated: `bet` is deprecated, use `beth` instead
 bet ב
-gimmel ג
 gimel ג
+@deprecated: `gimmel` is deprecated, use `gimel` instead
+ gimmel ג
```
mkorje, line 1061:
```suggestion
gimmel ג
```

mkorje (approved):

MDLC01:
I just realized that we should have waited for @laurmaedje's approval before merging this PR, as it plans a breaking change.

PR #63, opened 2025-05-17
Remove chinese compatibility characters

Enivex:
Remove the `small` variants of `ast`, `plus`, `eq`, `gt` and `lt`.

These are compatibility symbols for Chinese in the [Small Form Variants](https://en.wikipedia.org/wiki/Small_Form_Variants) block, and should not have been added in the first place.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -103,7 +103,6 @@ ast
   .low ⁎
   .double ⁑
   .triple ⁂
-  .small ﹡
   .circle ⊛
   .square ⧆
 at @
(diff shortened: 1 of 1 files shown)


MDLC01 (approved):

mkorje:
Should be wait until we have a way to properly deprecate modifiers?

MDLC01:
I think in this case it's fine because they are likely not widely used. Also, in this case, the deprecation would just say "this variant is deprecated, with no alternative" so it's not very useful.

mkorje (approved):

MDLC01:
I just realized that we should have waited for @laurmaedje's approval before merging this PR, as it contans a breaking change.

PR #65, opened 2025-05-18
Add astronomical symbols

Enivex:
Adds various astronomical symbols. These days, I believe only the symbols for the sun, earth and jupiter are commonly used in actual science, but do see use in astrology.

There's also
- ♇ for Pluto, but I skipped that one since it's no longer a planet.
- A bunch of moon symbols, some which can act as emojis

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -550,6 +550,20 @@ parallel ∥
 perp ⟂
   .circle ⦹
 
+// Astronomical.
+earth 🜨
+  .alt ♁
+jupiter ♃
+mars ♂
+mercury ☿
+neptune ♆
+  .alt ⯉
+saturn ♄
+sun ☉
+uranus ⛢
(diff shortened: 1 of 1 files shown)


MDLC01:
If Pluto is important in astrology, maybe we should add its symbol even if astronomers no longer consider it a planet?

Enivex:
> If Pluto is important in astrology, maybe we should add its symbol even if astronomers no longer consider it a planet?

Pluto has a separate symbol used for astrology apparently.

There's a bunch more symbols that could be added, for dwarf planets and such. The question is whether that should go in this PR or a separate one.

MDLC01:
If there are more symbols, let's keep them for a future PR.

MDLC01 (approved):

T0mstone (approved):

mkorje (approved):

PR #66, opened 2025-05-18
Add angzarr

Enivex:
We were missing the most important unicode symbol.

Context: https://en.wikipedia.org/wiki/Angzarr

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -537,6 +537,7 @@ divides ∣
 wreath ≀
 
 // Geometry.
+angzarr ⍼
 parallel ∥
   .struck ⫲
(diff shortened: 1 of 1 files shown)


MDLC01 (approved):
As an unused symbol, it may be a bit useless to add it. But at the same time it is quite well-known and so I believe it makes sense to add it either way because people may want to write _about_ it even if no one _uses_ it.

Btw, it appears we now know the original meaning of ⍼. But since no one uses it in that way anymore, I believe it makes more sense to add it under its ghost character name, under which it is most commonly known, rather than under a semantic name that corresponds to a semantic it is no longer used for.

mkorje (approved):

PR #68, opened 2025-05-19
Rename angle brackets and quotes to chevron

Enivex:
Currently, we had two entirely different families of symbols under `angle`. Namely all the angle brackets, and the geometry kind. I've renamed the brackets to chevron, which is an alternate name for them that someone suggested. At the same time, I moved the remaining angles to the "Geometry" section.

For consistency, I also renamed `quote.angle` to `quote.chevron`.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -63,7 +63,7 @@ fence
   .r ⧙
   .r.double ⧛
   .dotted ⦙
-angle ∠
+chevron
   .l ⟨
   .l.curly ⧼
   .l.dot ⦑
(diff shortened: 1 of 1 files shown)


T0mstone:
I'm not a big fan of `chevron`. There's also already U+1FBCA 🯊 WHITE UP-POINTING CHEVRON in Unicode, which is something entirely different.

Thinking of other shapes with an obtuse angle, I came up with "roof", but it's not really a good name either (especially given that these roofs (rooves?) would be sideways)...

Enivex:
> I'm not a big fan of `chevron`. There's also already U+1FBCA 🯊 WHITE UP-POINTING CHEVRON in Unicode, which is something entirely different.

Well, I wouldn't say entirely different. In any case, that symbol is in the "legacy computing" block of unicode, and therefore highly specialized. I don't think there's any danger in creating conflicts here.

I didn't take it out of thin air by the way. They can literally be called chevrons.

See e.g https://en.wikipedia.org/wiki/Bracket#Angle_brackets and https://en.wiktionary.org/wiki/chevron (meaning 5.)

MDLC01:
To make `chevron` callable, it would have to be added here:

https://github.com/typst/typst/blob/df89a0e85b80844ef56a6fa98af01eaaf7553da8/crates/typst-library/src/foundations/symbol.rs#L101-L126

MDLC01:
With https://github.com/typst/codex/pull/86 merged, this can now use the dedicated deprecation syntax for variants.

Enivex:
> With #86 merged, this can now use the dedicated deprecation syntax for variants.

Done

knuesel:
I'm in favor: it makes much sense to distinguish these two families and I think "chevron" is the best name we found.

MDLC01:
I don't remember if I expressed my opinion on that already, but I am in favor of this change for same reasons as @knuesel mentioned above.

However, I am wondering whether the best default for `chevron` is the angle brackets, or the angle quotes.

knuesel (approved):

PR #69, opened 2025-05-19
Add most missing delimiters (and more)

Enivex:
Adds most of the missing delimiters. These are still missing, mostly because I struggle with naming:
⦌  right square bracket with underbar (U+298C) 
⦔ right arc greater-than bracket (U+2994)
⦖ double right arc less-than bracket (U+2996) 

The first one would be easier to name once we have a name for the "undermacron".

The proposal document also lists double parentheses, but those do not seem to be math delimiters. I'm not sure what they are exactly, but they're not present in any font I tried.

- I renamed some of the `double` variants to `stroked` for consistency. This is especially important for the tortoise shells, since there is a filled variant.
- I hope the distinction between `bar` (meant to indicate the delimiter) and `stroked` is clear enough.
- I moved `amp` to punctuation, as it was in the wrong place.
- There is some weirdness regarding `fence`. `fence.dotted` is not a delimiter, and seems entirely unrelated. I didn't do anything with it, however.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -27,24 +27,24 @@ paren
   .l.flat ⟮
   .l.bar ⦇
   .l.stroked ⦅
-  //Deprecated, use paren.l.stroked
+  @deprecated: `paren.l.double` is deprecated, use `paren.l.stroked` instead
   .l.double ⦅
   .r )
   .r.flat ⟯
   .r.bar ⦈
   .r.stroked ⦆
-  //Deprecated, use paren.r.stroked
+  @deprecated: `paren.r.double` is deprecated, use `paren.r.stroked` instead
   .r.double ⦆
   .t ⏜
   .b ⏝
 brace
   .l U+7B
   .l.stroked ⦃
-  //Deprecated, use brace.l.stroked
+  @deprecated: `brace.l.double` is deprecated, use `brace.l.stroked` instead
   .l.double ⦃
   .r U+7D
   .r.stroked ⦄
-  //Deprecated, use brace.r.stroked
+  @deprecated: `brace.r.double` is deprecated, use `brace.r.stroked` instead
   .r.double ⦄
   .t ⏞
   .b ⏟
@@ -53,13 +53,13 @@ bracket
   .l.tick.t ⦍
   .l.tick.b ⦏
   .l.stroked ⟦
-  //Deprecated, use bracket.l.stroked
+  @deprecated: `bracket.l.double` is deprecated, use `bracket.l.stroked` instead
   .l.double ⟦
   .r ]
   .r.tick.t ⦐
   .r.tick.b ⦎
   .r.stroked ⟧
-  //Deprecated, use bracket.r.stroked
+  @deprecated: `bracket.r.double` is deprecated, use `bracket.r.stroked` instead
   .r.double ⟧
   .t ⎴
   .b ⎵
@@ -67,12 +67,12 @@ shell
   .l ❲
   .l.stroked ⟬
   .l.filled ⦗
-  //Deprecated, use shell.l.stroked
+  @deprecated: `shell.l.double` is deprecated, use `shell.l.stroked` instead
   .l.double ⟬
   .r ❳
   .r.stroked ⟭
   .r.filled ⦘
-  //Deprecated, use shell.r.stroked
+  @deprecated: `shell.r.double` is deprecated, use `shell.r.stroked` instead
   .r.double ⟭
   .t ⏠
   .b ⏡
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -24,32 +24,64 @@ space U+20
 // Delimiters.
 paren
   .l (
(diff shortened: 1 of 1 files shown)


src/modules/sym.txt
```diff
@@ -63,14 +95,21 @@ fence
   .r ⧙
   .r.double ⧛
   .dotted ⦙
+corner
+  .l.t ⌜
+  .l.b ⌞
+  .r.t ⌝
+  .r.b ⌟
```
MDLC01, line 102:
```suggestion
  .tl ⌜
  .bl ⌞
  .tr ⌝
  .br ⌟
```

src/modules/sym.txt
```diff
@@ -63,14 +95,21 @@ fence
...
 angle ∠
   .l ⟨
   .l.curly ⧼
   .l.dot ⦑
+  .l.bar ⦉
```
MDLC01, line 107: If #68 is accepted, this should become a variant of `chevron`.

src/modules/sym.txt
```diff
@@ -63,14 +95,21 @@ fence
...
   .l.double ⟪
   .r ⟩
   .r.curly ⧽
   .r.dot ⦒
+  .r.bar ⦊
```
MDLC01, line 112: If #68 is accepted, this should become a variant of `chevron`.

src/modules/sym.txt
MDLC01, line 107: I don't like the use of `.bar` here, which is typically for bars that appear over or under a symbol (except for arrows). Maybe `.closed` would be better?

MDLC01 (changes requested):
I strongly dislike the change from `.double` to `stroked`. When I talk about ⟦ and ⟧ out loud, I refer to them as "double brackets" or "double square brackets", not "stroked brackets". I understand that this makes sense for shells, but really, even `shell.{l,r}.filled` sounds bad imo. If we can't find a better solution for now, I suggest we keep `.double` instead of `.stroked`, and not introduce `shell.{l,r}.filled` for now. I am not aware of any use for them, it's not like it's a symbol we _need_ to add immediately.

src/modules/sym.txt
Enivex: Bar is meant to refer to the delimiter here

src/modules/sym.txt
Enivex: I don't think I agree with this change. It's pretty important here that the `l` modifier means a left delimiter and `r` means a right delimiter.

T0mstone: Tho with `.l.t` there's the issue of writing `corner.l` or `corner.t`. This could be resolved with the minimum modifier set idea, but currently the issue would be there.

Enivex:
> I strongly dislike the change from `.double` to `stroked`. When I talk about ⟦ and ⟧ out loud, I refer to them as "double brackets" or "double square brackets", not "stroked brackets". I understand that this makes sense for shells, but really, even `shell.{l,r}.filled` sounds bad imo. If we can't find a better solution for now, I suggest we keep `.double` instead of `.stroked`, and not introduce `shell.{l,r}.filled` for now. I am not aware of any use for them, it's not like it's a symbol we _need_ to add immediately.

The thing is that we're using `double` in an inconsistent way. Everywhere else it means two of something, but that's not what this is. I kept the `double` modifier for angle brackets and bars, because there the symbol is just that. For the other delimiters, it's literally the same kind of thing that we call stroked for other symbols. The unicode name for "double brackets" is "mathematical left white square bracket", and "white" is basically the unicode name for our stroked modifier.

I highly doubt that anyone is going to use these directly without defining some shorthand for them, so I don't see it as a big problem that the name isn't exactly what you would say out loud.

src/modules/sym.txt
Enivex: I may add that ⧞ is infinity.bar

MDLC01: I don't this ⧞ is the mos intuitive result for `infinity.bar`, and i still believe `angle.{l,r}.closed` would be better (or rather, `chevron.{l,r}.closed`, taking https://github.com/typst/codex/pull/68  into account).

Enivex: Both ⦇ and ⦅ are closed, but are distinct symbols

MDLC01: The first one comes from Z notation (⦉ as well btw), which I don't think we have named symbols for as of now.[^1] I'm wondering whether we should add them. Although I guess I see no immediate reason not to.

[^1]: Only exception I could find for Typst 0.13.1: `circle.filled.tiny` for U+2981 ⦁ Z Notation Spot.

Enivex: I see no reason not to add them at some point, but that seems out of scope for a PR focused on delimiters

MDLC01:
With https://github.com/typst/codex/pull/86 merged, this can now use the dedicated deprecation syntax for variants.

Enivex:
I'll create new smaller PRs. This one was too large.

PR #74, opened 2025-05-22
Add playing cards

Enivex:
- One possibility is moving the suit symbols into this, but that would be a breaking change.
- I considered using roman numerals for the the numbering of the trump cards. However, we've just spelled out numbers elsewhere, so I did that for consistency.
- Each tarot card also has a meaning (https://en.wikipedia.org/wiki/Bourgeois_Tarot#Tarot_Nouveau) which could possibly be an alias, but I'm not sure if it's worth it.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -616,6 +616,97 @@ suit
   .spade.filled ♠
   .spade.stroked ♤
 
+card {
+  spade
+    .ace 🂡
+    .two 🂢
+    .three 🂣
+    .four 🂤
+    .five 🂥
+    .six 🂦
+    .seven 🂧
+    .eight 🂨
+    .nine 🂩
+    .ten 🂪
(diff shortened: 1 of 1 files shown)


Enivex:
Hm, it's failing because of the text variation selector on `card.joker.black`

Enivex:
Also: Is `card` too generic? I couldn't think of another option not involving having to use a hyphen (which we've avoided so far)

T0mstone:
> Should I
> 
>    * Remove the variation selector
> 
>    * Remove the black joker for the time being
> 
>    * Wait until variation selectors are supported

I'd advocate for waiting on #25

Enivex:
It's a bit long, but presumably anyone using this would import the submodule anyway.

T0mstone (changes requested):

PR #75, opened 2025-05-24
Add domino tiles

Enivex:
A submodule turned out to be super useful here, because it made it possible to have order-dependent modifiers

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1132,6 +1132,117 @@ errorbar
   .circle.stroked ⧲
   .circle.filled ⧳
 
+domino {
+  back
+    .h 🀰
+    .v 🁢
+  zero
+    .zero.h 🀱
+    .zero.v 🁣
+    .one.h 🀲
+    .one.v 🁤
+    .two.h 🀳
+    .two.v 🁥
+    .three.h 🀴
+    .three.v 🁦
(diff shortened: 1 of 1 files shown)


MDLC01 (approved):

T0mstone:
There's a bit of a bikeshed here about whether the plain `n` domino should be `n.zero` or `n.n`.

Enivex:
I'm not well versed in dominoes, but it seems more natural to me that "five" refers to "five and nothing" instead of "five and five". The current order is also convenient because it's sorted.

Enivex:
Did you mean something like this then?
```
five
    .h 🁙
    .v 🂋
    .zero.h 🁔
    .zero.v 🂆
    .one.h 🁕
    .one.v 🂇
    .two.h 🁖
    .two.v 🂈
    .three.h 🁗
    .three.v 🂉
    .four.h 🁘
    .four.v 🂊
    .five.h 🁙
    .five.v 🂋
    .six.h 🁚
    .six.v 🂌
```

I'm not a fan, but I guess I'll defer to someone who actually knows anything about dominoes.

(There's also the question of whether the exact fallback is even important.)

Enivex:
Closing this until we've figured out https://github.com/typst/codex/issues/94 and https://github.com/typst/codex/issues/95

PR #76, opened 2025-05-24
Add mahjong tiles

Enivex:
Each suit has multiple different names. I went with the ones used in the official unicode names.

The red dragon tile needs a variation selector, so this cannot be merged yet.


T0mstone (changes requested):

PR #77, opened 2025-05-24
Add recycling symbols

Enivex:
The permanent paper sign is not technically a recycling symbol, but it's closely related and I'd not sure where else it'd go.

Two of the symbols need text variation selectors.

I went with "recycle" instead of "recycling" since it's shorter, but I can be convinced either way.


Enivex:
Closing until the multi-character symbols situation is resolved.

PR #79, opened 2025-05-31
Change some of the more incorrectly (?) named emojis

Enivex:
Current emoji naming is very inconsistent, and in some cases wrong. I initially started taking notes for every single emoji, but that turned out to simply be too much. This is a more constrained PR renaming some of the more clear cut cases as a start

We probably need some deprecation warnings if we want to proceed here, but I wanted to get feedback before adding them.

aesculapius -> asclepius: The former is the latin name. A more radical naming would be "medical"

aubergine -> eggplant: This is the name in US english

basecap -> cap: The former seems to be an example of https://en.wikipedia.org/wiki/Pseudo-anglicism . We could use "baseballcap" or "cap.baseball", but I don't think there are other cap emojis.

beetle.lady -> ladybug: This is the name in US english

bin -> wastebasket: Bin is rarely used in US english

camel.dromedar -> camel.dromedary: This was either a typo, or the German name.

cardindex -> rolodex: More common name, and it's literally what the emoji is depicting

cassette -> vhs: Cassette is more commonly used for audio cassettes, while the emoji depicts a video cassette (vhs).

clip -> paperclip: I'm not aware that the short form "clip" is used for this.
clips -> paperclip.linked: See above. An alternative is "paperclips", but the emoji does specifically depict linked paperclips.

crystal -> crystalball: I honestly think this might have just been an oversight?

cutlery -> silverware: Again, this is the most common terminology in US english

dino.pod -> dino.sauropod: The original name was strange. Alternatively "dino.saur" or just "dinosaur" or "sauropod"
dino.rex -> dino.trex: Alternatively just "trex"

feeding.breast -> breastfeeding

fuelpump -> gaspump: Common name in US english. Fuel pump can also mean other things.

helix -> dna: The emoji specifically depicts the DNA double helix.

moyai -> moai: typo?

partalteration -> partalternation: typo (as an aside, this seems very obscure)

railway -> railroad: US english

ringbuoy -> lifebuoy: US english

satdish -> satellite.dish

shoe.thong -> shoe.flipflop: US english

tm -> trademark: Alignment with symbols.txt
reg -> trademark.registered: Alignment with symbols.txt

envelope: This emoji was missing
tray.mail -> envelope.incoming: This seems to have been misplaced under tray.

unknown: Removed, since this was the "white hair" emoji component, and not an actual emoji.

### DIFF
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -391,7 +391,6 @@ diamond
   .orange.small 🔸
   .dot 💠
 die 🎲
-dino
 sauropod 🦕
 trex 🦖
 disc
@@ -1058,7 +1057,6 @@ placard 🪧
 planet 🪐
 plant 🪴
 plaster 🩹
-plate
 plate 🍽
 playback
   .down ⏬
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -1059,7 +1059,7 @@ planet 🪐
 plant 🪴
 plaster 🩹
 plate
-  .silverware 🍽
+plate 🍽
 playback
   .down ⏬
   .eject ⏏
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -195,7 +195,7 @@ calendar 📅
   .spiral 🗓
   .tearoff 📆
 camel 🐫
-  .dromedary 🐪
+dromedary 🐪
 camera 📷
   .flash 📸
   .movie 🎥
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -392,8 +392,8 @@ diamond
   .dot 💠
 die 🎲
 dino
-  .sauropod 🦕
-  .trex 🦖
+sauropod 🦕
+trex 🦖
 disc
   .cd 💿
   .dvd 📀
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -1143,7 +1143,7 @@ santa
   .man 🎅
   .woman 🤶
 satellite 🛰
-  .dish 📡
+antenna 📡
 saw 🪚
 saxophone 🎷
 scales ⚖
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -3,7 +3,7 @@ abc 🔤
 abcd 🔡
 ABCD 🔠
 accordion 🪗
-aesculapius ⚕
+asclepius ⚕
 airplane ✈
   .landing 🛬
   .small 🛩
@@ -49,7 +49,7 @@ ast *
   .box ✳
 atm 🏧
 atom ⚛
-aubergine 🍆
+eggplant 🍆
 avocado 🥑
 axe 🪓
 baby 👶
@@ -72,7 +72,7 @@ banjo 🪕
 bank 🏦
 barberpole 💈
 baseball ⚾
-basecap 🧢
+cap 🧢
 basket 🧺
 basketball ⛹
   .ball 🏀
@@ -95,7 +95,7 @@ beer 🍺
   .clink 🍻
 beet 🫜
 beetle 🪲
-  .lady 🐞
+ladybug 🐞
 bell 🔔
   .ding 🛎
   .not 🔕
@@ -106,7 +106,7 @@ bike 🚲
   .not 🚳
 bikini 👙
 billiards 🎱
-bin 🗑
+wastebasket 🗑
 biohazard ☣
 bird 🐦
 bison 🦬
@@ -195,7 +195,7 @@ calendar 📅
   .spiral 🗓
   .tearoff 📆
 camel 🐫
-  .dromedar 🐪
+  .dromedary 🐪
 camera 📷
   .flash 📸
   .movie 🎥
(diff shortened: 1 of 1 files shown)


Enivex:
In general, I'd say that modifiers are currently used both way too extensively, and not extensively enough.

Enivex:
This requires multi-character symbols because of the envelope emoji.

src/modules/emoji.txt
```diff
@@ -72,7 +72,7 @@ banjo 🪕
 bank 🏦
 barberpole 💈
 baseball ⚾
-basecap 🧢
+cap 🧢
```
T0mstone, line 75: `cap` might lead to slightly bad UX for people coming from LaTeX where `\cap` is our `inter`.

src/modules/emoji.txt
```diff
@@ -106,7 +106,7 @@ bike 🚲
   .not 🚳
 bikini 👙
 billiards 🎱
-bin 🗑
+wastebasket 🗑
```
T0mstone, line 109: This one seems kind of long, even tho it's the official Unicode name. I would suggest something like `trashbin`.

src/modules/emoji.txt
```diff
@@ -195,7 +195,7 @@ calendar 📅
   .spiral 🗓
   .tearoff 📆
 camel 🐫
-  .dromedar 🐪
```
T0mstone, line 198: (I can confirm that "Dromedar" is the German name)

T0mstone (commented):
General notes:
- You make frequent references to what emoji look like, but afaict this isn't actually standardized anywhere and entirely dependent on the font being used. Ideally, we should stick to what Unicode actually says.
- When, how and by whom was it decided that we would be using US english as a standard?

Enivex:
>     * You make frequent references to what emoji look like, but afaict this isn't actually standardized anywhere and entirely dependent on the font being used. Ideally, we should stick to what Unicode actually says.

No, there is no formal standardization, but major implementations have essentially converged on a common general look. This is important for e.g. interoperability reasons.

See https://www.unicode.org/reports/tr51/#Design_Guidelines

"While the shape of the character can vary significantly, designers should maintain the same “core” shape, based on the shapes used mostly commonly in industry practice"

As an example, see the entry for the card index emoji https://emojipedia.org/card-index#designs

>     * When, how and by whom was it decided that we would be using US english as a standard?

Just look at the language used in the rest of typst. It's clearly the US english spelling that is used. For instance "color" , "gray" and "transparentize", and not "colour", "grey" and "transparentise"

src/modules/emoji.txt
Enivex: Even if that were the case, I don't understand how this would be relevant when these are in the emoji namespace.

src/modules/emoji.txt
Enivex: There are a couple of points to make here. First, trash bin is a broader term than waste basket, which largely has the connotation of a receptacle for paper waste (think office). Second, bin is not commonly used in american english. We could use trash can instead, but that still has the first problem.

src/modules/emoji.txt
```diff
@@ -195,7 +195,7 @@ calendar 📅
...
+  .dromedary 🐪
```
MDLC01, line 198:
```suggestion
dromedary 🐪
```

Why make it a modifier in the first place?

src/modules/emoji.txt
```diff
@@ -392,8 +392,8 @@ diamond
   .dot 💠
 die 🎲
 dino
-  .pod 🦕
-  .rex 🦖
+  .sauropod 🦕
+  .trex 🦖
```
MDLC01, line 396:
```suggestion
sauropod 🦕
trex 🦖
```

I don't this this would be ambiguous.

src/modules/emoji.txt
```diff
@@ -1146,8 +1144,8 @@ sandwich 🥪
 santa
   .man 🎅
   .woman 🤶
-satdish 📡
 satellite 🛰
+  .dish 📡
```
MDLC01, line 1148: Not sure I like this being a variant of `satellite`. Maybe `antenna` or `antenna.dish` would be better?

src/modules/emoji.txt
```diff
@@ -1060,7 +1059,7 @@ planet 🪐
 plant 🪴
 plaster 🩹
 plate
-  .cutlery 🍽
+  .silverware 🍽
```
MDLC01, line 1062:
```suggestion
plate 🍽
```

There is no plate without silverware emoji AFAIK.

MDLC01 (commented):
I left some comments but did not think about everything into too much details.

src/modules/emoji.txt
Enivex: Makes sense

src/modules/emoji.txt
Enivex: I renamed it to just antenna.

MDLC01:
@laurmaedje this contains breaking changes

Enivex:
> @laurmaedje this contains breaking changes

Wouldn't be possible to merge it yet anyway due to the multi character symbol

T0mstone (changes requested):

PR #81, opened 2025-06-03
Deprecate kai and Kai

Enivex:
This is the greek version of ampersand, and wouldn't be useful for anyone not writing in greek.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1002,6 +1002,7 @@ epsilon ε
 eta η
 gamma γ
 iota ι
+@deprecated: `kai` is deprecated
 kai ϗ
 kappa κ
   .alt ϰ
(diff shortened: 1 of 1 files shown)


src/modules/sym.txt
```diff
@@ -1002,6 +1002,7 @@ epsilon ε
 eta η
 gamma γ
 iota ι
+@deprecated: `kai` is deprecated
```
laurmaedje, line 1005: it would be good if the deprecation message stated, what to do / use instead. In this case, that's just entering the symbol directly I guess, but it's still useful to know that that's the intended thing to do.

Enivex: That's fair, though I can't imagine many people will see this deprecation message.

MDLC01 (approved):

Enivex:
I improved the deprecation messages. Though I'm not sure if any additional parts of the messages need to be inside backticks. Particularly the backslash,.

mkorje (approved):

PR #82, opened 2025-06-03
Add digamma

Enivex:
This is an archaic greek letter, but it does see use (though rarely), and appears in reasonably complete math fonts. It's also the only one that gets a name in unicode-math (the latex package).

Other archaic symbols exist, but I don't think there is a point in adding them unless requested.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -997,6 +997,7 @@ beta β
   .alt ϐ
 chi χ
 delta δ
+digamma ϝ
 epsilon ε
   .alt ϵ
 eta η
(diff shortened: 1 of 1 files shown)


mkorje (approved):
digamma has also got a bold mathematical variant (already supported in Typst), so I think it makes sense to add it.

MDLC01 (approved):
For context regarding uses for capital Digamma:
- https://en.wikipedia.org/wiki/Digamma_function
- https://en.wikipedia.org/wiki/750_GeV_diphoton_excess

MDLC01:
I agree we shouldn't add other archaic Greek letters unless requested, at least for now.

Enivex:
> For context regarding uses for capital Digamma:
> - https://en.wikipedia.org/wiki/Digamma_function
> - https://en.wikipedia.org/wiki/750_GeV_diphoton_excess

Ironically the digamma function is rarely denoted by digamma

PR #83, opened 2025-06-03
Add Theta.alt

Enivex:
I happened to notice in https://us.mirrors.cicku.me/ctan/macros/unicodetex/latex/unicode-math/unimath-symbols.pdf that a single capital Greek letter has an alternate version, and that we were missing it.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1050,6 +1050,7 @@ Rho Ρ
 Sigma Σ
 Tau Τ
 Theta Θ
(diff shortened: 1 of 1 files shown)


mkorje (approved):

Enivex:
I'm pretty sure it's just an alternate. There are even bold and italic versions of it.

It's not _as_ distinct as some versions, but it's not terribly hard to tell the difference either. I think replacing them would be a mistake.

PR #86, opened 2025-06-10
Allow deprecating variants

MDLC01:
This allows deprecating variants, which I did for the two variants that are already deprecated. I will start working on a corresponding typst/typst PR soon. I may wait until https://github.com/typst/typst/pull/6159 is merged first.

With the current system, there is a weird edge case where modifierless variants can be deprecated independently of the symbol itself, although there is no syntactical way to deprecate a modifierless variant in out DSL. I don't think this is too much of an issue. It could even prove useful if we want to change the default variant of symbol (first, mark the modifierless variant as deprecated, then, in a future version, change its value and remove the deprecation), although this will require adding new syntax at this point.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -31,10 +31,10 @@ enum Def<'a> {
     Module(Module<'a>),
 }
 
-/// A symbol, either a leaf or with modifiers.
+/// A symbol, either a leaf or with modifiers with optional deprecation.
 enum Symbol<'a> {
     Single(char),
-    Multi(Vec<(ModifierSet<&'a str>, char)>),
+    Multi(Vec<(ModifierSet<&'a str>, char, Option<&'a str>)>),
 }
 
 /// A single line during parsing.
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -66,11 +74,33 @@ fn process(buf: &mut String, file: &Path, name: &str, desc: &str) {
...
+                    Some(Err(String::from("duplicate `@deprecated:`")))
+                } else {
+                    deprecation = Some(message);
+                    None
+                }
+            }
+            Ok(Line::ModuleStart(name)) => {
+                Some(Ok(Declaration::ModuleStart(name, deprecation.take())))
+            }
+            Ok(Line::ModuleEnd) => Some(Ok(Declaration::ModuleEnd)),
```
T0mstone, line 96: This should error if `deprecation.is_some()`, just like the previous code did.

build.rs
```diff
@@ -66,11 +74,33 @@ fn process(buf: &mut String, file: &Path, name: &str, desc: &str) {
...
+            Ok(Line::Symbol(name, c)) => {
+                Some(Ok(Declaration::Symbol(name, c, deprecation.take())))
+            }
+            Ok(Line::Variant(modifiers, c)) => {
+                Some(Ok(Declaration::Variant(modifiers, c, deprecation.take())))
+            }
+        })
```
T0mstone, line 103: The previous code also ensured that the iterator can't end with a deprecation. Maybe this could be achieved by adding `.chain(once_with(|| /* check */).filter_map(|_| None))` here?

build.rs
MDLC01: Good catch!

laurmaedje (approved):

MDLC01:
Will get to it tonight

PR #87, opened 2025-06-12
Add `eq.dot.up` and `eq.dot.down`

cAttte:
hi, this PR adds the symbols U+2250 Approaches the Limit ([thanks](https://discord.com/channels/1054443721975922748/1088371919725793360/1381955865086791740)) and U+2A66 Equals Sign with Dot Below as `eq.dot.up` and `eq.dot.down`. not sure if the names are right (not a lot of precedent). also added the `up` variant first because it's probably what a user is expecting when they type `eq.dot`, and it's analogous to LaTeX doteq

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -272,6 +272,8 @@ eq =
   .star ≛
   .circle ⊜
   .colon ≕
+  .dot.up ≐
+  .dot.down ⩦
   .dots ≑
   .dots.down ≒
   .dots.up ≓


MDLC01:
`.up` and `.down` are used for rising and falling dots. As you noted, there is no prior art, but this was briefly discussed in https://github.com/typst/typst/pull/4939. I am personally in favor of using `.above` and `.below`.

Enivex:
This is covered by @MDLC01 's proposal document. There's a number of these that need to be added. The holdup is largely that there needs to be consensus on naming of the modifier.

I think the main options are `above` and `below`, or `over` and `under`.

Enivex:
I think it makes sense to close this PR until we can discuss naming of directions and positions more broadly. You are of course welcome to contribute.

Issue #42, opened 2025-01-22
Use shorter modifier names for slanted equal


MDLC01:
## Context

`lt.eq.slant` and `gt.eq.slant` are very long names for quite common symbols. It is also not immediately clear that the `.slant` modifier applies to the equal part of the symbol (`lt.slant.eq` and `gt.slant.eq` are equally valid, but ambiguous). For more context, see [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1285615254440644771).

Similarly, `prec.curly.eq` (⪯) and `succ.curly.eq` (≽) make no sense.

In the following, I replace `.eq.slant` on `lt` and `gt`, and `.curly.eq` on `prec` and `succ`, with `.seq`. This is intended to be debated in this issue's comment thread.

## Formal proposal

Rename four symbols from `sym` according to the following table.

| Character | Current name | New name |
|------------|----------------|-------------|
| ⩽ LESS-THAN OR SLANTED EQUAL TO | `lt.eq.slant` | `lt.seq` |
| ⩾ GREATER-THAN OR SLANTED EQUAL TO | `gt.eq.slant` | `gt.seq` |
| ≼ PRECEDES OR EQUAL TO | `prec.curly.eq` | `prec.seq` |
| ≽ SUCCEEDS OR EQUAL TO | `succ.curly.eq` | `succ.seq` |

## Breaking changes

`lt.eq.slant` and `gt.eq.slant` would become invalid.

## Notable information

Unicode defines the characters ⋞ EQUAL TO OR PRECEDES and ⋟ EQUAL TO OR SUCCEEDS, but no variant with a single horizontal line equal sign above ≺ or ≻. This may indicate that ≼ should be `prec.eq`, and ⪯ should be considered less common, and therefore use a more complex variant than plain `eq` (and similarly for ≽ and ⪰).

Issue #51, opened 2025-02-19
Improve forward compatibility with a notion of minimal modifier set


MDLC01:
## Motivation

Currently, when selecting a variant from a set of modifiers, the first variant from the list that contains all the modifiers, and a minimal amount of additional modifiers, is chosen.[^1][^2] This means using non-fully qualified names when referring to a symbol might cause breakage when Codex is updated. For example, consider the following symbol:

```
arrow
  .l ←
  .r →
  .r.bar ↦
```

`arrow.bar` resolves to `arrow.r.bar`, which is ↦. Now, suppose a new version of Codex changes the symbol to the following:

```
arrow
  .l ←
  .l.bar ↤
  .r →
  .r.bar ↦
```

Now, `arrow.bar` will resolve to `arrow.l.bar`, which is ↤.

Essentially, this means adding new variants in the middle of the variant list can cause unexpected breakage. As of now, there is no policy regarding what constitute a breaking change when it comes to fallbacks.

[^1]: This is actually not defined in Codex, but in Typst. Making the variant selection part of Codex is the topic of #30.
[^2]: https://github.com/typst/typst/blob/d199546f9fe92b2d380dc337298fdca3e6fca8c8/crates/typst-library/src/foundations/symbol.rs#L387-L420

## Proposed solution

The intuitive idea of this solution is to make explicit what parts of the fully qualified variant (i.e., which modifiers) can be omitted.

As before, each variant has a set of modifiers. Hereafter, we refer to this set of modifiers as the _fully qualified variant_, denoted $M_\text{full}$. Additionally, a variant can define a _minimal modifier set_, denoted $M_\text{min}$, which is a subset of the fully qualified variant (i.e., $M_\text{min} \subseteq M_\text{full}$). When selecting a variant for a set of modifiers $M$, the same process as before is applied, with the additional constraint that the selected variant's minimal modifier set is included in $M$ (i.e., $M_\text{min} \subseteq M$).

The current behavior corresponds to having $M_\text{min} = \emptyset$ for all variants. With this proposal, the default would become $M_\text{min} = M_\text{full}$, with lots of manual overrides to allow for backward compatibility and more leniency.

This improves forward-compatibility by explicitly specifying which fallbacks can be relied on, and which can't. Ideally, there could even be an automated way of detecting breakages. This is currently not feasible, because most breakages are not implicitly guaranteed to be future proof.[^3]

[...]

PR #89, opened 2025-06-13
Implement shorter modifier name for slanted equal and equiv

Enivex:
This implements the changes described in #42. Also
- Add additional related variants that should be uncontroversial based on existing choices. There are some more in the document ™, but they require more discussion.
- Naturally introduce the corresponding `sequiv`.
- This should unblock https://github.com/typst/codex/pull/36, since curly is now freed up

While there a few concerns that have been voiced, I've also not seen any viable alternative for resolving ambiguity. Even `seq` is not enough to resolve ambiguity for some of the more cursed symbols (⪓).

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -289,6 +289,11 @@ eq =
   .triple ≡
   .triple.not ≢
   .quad ≣
+seq
+  .lt ⪕
+  .lt.dot ⪗
+  .gt ⪖
+  .gt.dot ⪘
 gt >
   .circle ⧁
   .dot ⋗
@@ -296,6 +301,7 @@ gt >
   .double ≫
   .eq ≥
   .seq ⩾
+  .dot.seq ⪀
   @deprecated: `gt.eq.slant` is deprecated, use `gt.seq` instead
   .eq.slant ⩾
   .eq.lt ⋛
@@ -323,6 +329,7 @@ lt <
   .double ≪
   .eq ≤
   .seq ⩽
+  .dot.seq ⩿
   @deprecated: `lt.eq.slant` is deprecated, use `lt.seq` instead
   .eq.slant ⩽
   .eq.gt ⋚
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -295,6 +295,8 @@ gt >
   .approx ⪆
   .double ≫
   .eq ≥
+  .seq ⩾
+  @deprecated: `gt.eq.slant` is deprecated, use `gt.seq` instead
   .eq.slant ⩾
   .eq.lt ⋛
   .eq.not ≱
@@ -320,6 +322,8 @@ lt <
   .approx ⪅
   .double ≪
   .eq ≤
+  .seq ⩽
+  @deprecated: `lt.eq.slant` is deprecated, use `lt.seq` instead
   .eq.slant ⩽
   .eq.gt ⋚
   .eq.not ≰
@@ -344,7 +348,11 @@ approx ≈
   .not ≉
 prec ≺
   .approx ⪷
+  .seq ≼
+  @deprecated: `prec.curly.eq` is deprecated, use `prec.seq` instead
   .curly.eq ≼
+  .seq.not ⋠
+  @deprecated: `prec.curly.eq.not` is deprecated, use `prec.seq.not` instead
   .curly.eq.not ⋠
   .double ⪻
   .eq ⪯
@@ -357,7 +365,11 @@ prec ≺
   .tilde ≾
 succ ≻
   .approx ⪸
+  .seq ≽
+  @deprecated: `prec.curly.eq` is deprecated, use `prec.seq` instead
   .curly.eq ≽
+  .seq.not ⋡
+  @deprecated: `prec.curly.eq.not` is deprecated, use `prec.seq.not` instead
   .curly.eq.not ⋡
   .double ⪼
   .eq ⪰
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -307,6 +307,7 @@ gt >
   .eq.lt ⋛
   .eq.not ≱
   .equiv ≧
+  .sequiv ⫺
   .lt ≷
   .lt.not ≹
   .neq ⪈
@@ -335,6 +336,7 @@ lt <
   .eq.gt ⋚
   .eq.not ≰
   .equiv ≦
+  .sequiv ⫹
   .gt ≶
   .gt.not ≸
   .neq ⪇
@@ -389,6 +391,9 @@ succ ≻
   .tilde ≿
 equiv ≡
   .not ≢
+sequiv
+  .lt ⪛
+  .gt ⪜
 smt ⪪
   .eq ⪬
 lat ⪫


MDLC01:
I don't love the `seq` and `sequiv` names, but I agree we don't have anything better.

However, I strongly dislike having `seq` and `sequiv` as separate symbols, because `seq` and `sequiv` don't make sense by themselves.

Maybe a solution would be to replace `.slant` with `.alt`. It's not very descriptive, but it may be the right solution. After all, slanted variants are mostly alternate glyphs of non-slanted variants, just like `epsilon.alt` and `phi.alt` are alternate glyphs for `epsilon` and `phi`.

Enivex:
> I don't love the `seq` and `sequiv` names, but I agree we don't have anything better.
> 
> However, I strongly dislike having `seq` and `sequiv` as separate symbols, because `seq` and `sequiv` don't make sense by themselves.

I don't see this as an issue at all. It just makes them usable in the same way the `eq` variants are. There are plenty of examples of symbols where there is no inherently canonical top variant.

We can be super principled, but ultimately this is a tool we want people to actually use.

It only makes sense to have `seq.lt` and `seq.gt` as counterpoints to `lt.seq` and `gt.seq`.
 
> Maybe a solution would be to replace `.slant` with `.alt`. It's not very descriptive, but it may be the right solution. After all, slanted variants are mostly alternate glyphs of non-slanted variants, just like `epsilon.alt` and `phi.alt` are alternate glyphs for `epsilon` and `phi`.

This has the same issues as `slanted`, but even more so. What does the `alt` refer to?

Enivex:
> 1. `.alt` provides an alternate version of the whole symbol. In this case, this corresponds to a slanted equal/equiv part.

But how does the user know what's going to change?

>     4. Does this mean ⪨ would be `whatever.seq` instead of `whatever.eq`? With my `.alt` proposal, it would simply be `whatever.eq` as there is no alternative variant here (unless there is with variation selectors, in which case we should add both).

Yes. Using `whatever.eq` would be both inconsistent and possibly lead to breaking changes down the line if a straight equality version is added. (It has happened several times before that "missing" symbol variants have been filled in)

>     5. Except, e.g., `lt.neq` is not an alternate of `lt.eq`, it's a different symbol with a different meaning. On the contrary, the `.alt` modifier does not change the meaning of the symbol, but uses an alternate representation of the symbol.

The slanted equal can also have a different meaning, such as inequality modulo a multiplicative constant. It depends on the author.

Enivex:
> This is a fair point, but I don't think it is a huge problem: they can just try. Same thing with `epsilon.alt` which could be `epsilon.lunate` The modifier does not say "this is what changes", but it provides a common variant of the symbol. I think this is fine, not every detail needs to be encoded in the identifier.

- The alternate greek symbol variants are established and well known (though a couple are reversed in typst)
- With the exception of `note.quarter.alt`,  `note.eighth.alt` (which I'm not a fan of either) and `epsilon.alt.rev` (and only because we reversed epsilon), every single use of `.alt` is as a single modifier, not as an alternate version of a symbol with a bunch of other modifiers.
- Someone writing `gt.alt` or `eq.alt` will be very confused.

> > Yes. Using `whatever.eq` would be both inconsistent
> 
> I don't think so. In my mind, `.eq` roughly means "or equal to", not "or equal to but horizontal".

In my mind it means the latter, because that's how it's consistently used across the board. With the exception of the symbols we're trying to fix now.

> > and possibly lead to breaking changes down the line if a straight equality version is added. (It has happened several times before that "missing" symbol variants have been filled in)
> 
> This is a fair point. However, I think Unicode could break many names if they decided to add the right characters. We can't think just in terms of what could be there.

No, but we don't have to make the job harder for ourselves for no reason either.
 
> > The slanted equal can also have a different meaning, such as inequality modulo a multiplicative constant. It depends on the author.
> 
> I did not know that. Do you have examples of documents that use both symbols with different meanings?

I don't have any explicit examples no, but I've seen both ≲ ≺ and ⩽ used for this purpose, with the first one being by far the most common.

MDLC01:
> The alternate greek symbol variants are established and well known (though a couple are reversed in typst)

I think the same could be said of the two variants of the less than or equal signs.

> every single use of `.alt` is as a single modifier, not as an alternate version of a symbol with a bunch of other modifiers.

I don't see why we would want to keep that as an invariant of symbol naming.

> Someone writing `gt.alt` or `eq.alt` will be very confused.

This is a fair point, but also a more general problem with our current system. https://github.com/typst/codex/issues/51 aims to solve that, but before this or a different solution is implemented, I agree this would be a bit awkward (although not much more than any other similar situation that already exists).

> In my mind it means the latter

Fair.

> No, but we don't have to make the job harder for ourselves for no reason either.

Fair.

> I don't have any explicit examples no, but I've seen both ≲ ≺ and ⩽ used for this purpose, with the first one being by far the most common.

Without actual examples of ⩽ and ≤ being used with different meanings, I can't help but think of both as alternate glyphs of each other.

MDLC01:
> What about `lt.eq.t` for `⋜` and `lt.seq.t` for `⪕`?

It is ambiguous what `t` refers to in those.

To go in the direction of https://github.com/typst/codex/pull/89#issuecomment-3024515803, I would say that modifiers that correspond to symbols can mean one of three things in the current system (and this is not a problem imo):
- The symbol is added around the base like an accent or an attachment (smaller than the base).
- The symbol is stacked below the base.
- The symbol is stacked to the right of the base.
This is an existing convention and I think it works quite well for the symbols that we currently have.

knuesel:
That makes sense. I don't have any objection then.

(I agree with @T0mstone that `seq.gt` feels wrong, but so does the symbol, so it fits :-) )

T0mstone:
If everyone is in consensus against me, I would at least want `seq` on its own to become invalid if we merge #105.

Enivex:
> If everyone is in consensus against me, I would at least want `seq` on its own to become invalid if we merge #105.

That's reasonable

T0mstone:
> I would like to know your opinions on using `.alt` as suggested in [#89 (comment)](https://github.com/typst/codex/pull/89#issuecomment-2971634604). I think only Enivex had something to say about it.

I like `.alt` more than `.seq`, but I agree with Enivex that it's a bit unfortunate if that also gives us `lt.alt`. However, if we merge #105, then I'd be fully in favor of `.alt` with `.eq` being required for the variant.

As someone who usually prefers `lt.eq.alt` over `lt.eq`, I often use a show rule `show lt.eq: lt.eq.slanted` so ease of access isn't really a concern for me personally, tho I have to concede that that's not an option for everyone (e.g. if you also have to use the non-slanted version sometimes). But besides that, I also feel like `lt.eq.alt` is conceptually more "correct" and fits in perfectly with the already existing uses of `.alt`.

Enivex:
To be clear on my stance

- I prefer `.slant` over `.slanted`, because the `ed` is completely unnecessary verbosity.
- I prefer both `.slant` and `.slanted` over `.alt`
-  I strongly prefer using `.seq` and `.sequiv` over all of them.
  - In usage it's completely consistent with the pattern of how every other relation works.
  - It makes the symbols usable without show rules
  - It's clear which part of the symbol it applies to (just like for `.neq`)
  - I think the worry about `seq` making sense on its own is unwarranted. I doubt it would even cross any users mind. On the other hand, they would be more confused if there wasn't a corresponding `seq.lt` to `lt.seq`.

MDLC01:
Superseding PR: https://github.com/typst/codex/pull/111

PR #90, opened 2025-06-13
Add and deprecate post 0.13.1 removals

Enivex:


Enivex:
> Are there other symbols that were removed after the last release or is this it?

As far as I know, this is it.

MDLC01:
> I wasn't sure what to write for the compatibility characters. Most users should not be using them at all.

I think this is fine. A whole explanation would be too long imo for a deprecation message.

T0mstone:
> Are there other symbols that were removed after the last release or is this it?

0.11.1 is the fourth-to-last release (if I counted correctly).

Enivex:
> > Are there other symbols that were removed after the last release or is this it?
> 
> 0.11.1 is the fourth-to-last release (if I counted correctly).

Oops, I obviously meant 0.13.1

mkorje (approved):

PR #91, opened 2025-06-14
Add bullets

### LINKED ISSUE #88: Triangular bullet
If I'm not mistaken, ‣, *U+2023 TRIANGULAR BULLET* which is `list.marker.at(1)` by default in Typst 0.13.1, does not have a name as a symbol. It would be good to have it available for easy list configuration, like `set list(marker: sym.bullet.tri)` (?)

Enivex:
There are also some additional dingbats, but I avoided adding them both because naming is hard, they double as emojis, and it's not entirely clear whether they belong under bullet or not.

Closes #88

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -683,6 +683,15 @@ sharp ♯
 
 // Shapes.
 bullet •
+  .op ∙
+  .o ⦿
+  .stroked ◦
+  .stroked.o ⦾
+  .hole ◘
+  .hyph ⁃
+  .tri ‣
+  .l ⁌
+  .r ⁍
 circle
   .stroked ○
   .stroked.tiny ∘


MDLC01:
U+2022 • BULLET (`sym.bullet`) has the Unicode informative alias "black small circle", but is not present in Typst as a variant of `sym.circle`. U+2219 ∙ BULLET OPERATOR (`sym.bullet.op` in this PR) has no Unicode informative alias, but is already present as `sym.circle.filled.small`.

Do you know what the point of `bullet.hole` is?

MDLC01 (approved):
Seeing as it seems to be used as list markers in some documents, let's keep `bullet.hole`.

T0mstone (approved):
Similarly to `circle.stroked.tiny` with `compose.o`, we'll probably want to deprecate `circle.filled.small` in favor of `bullet.op` too. (Both of these codepoints are in the Mathematical Operators block)

PR #92, opened 2025-06-23
Allow multi-character symbols/variants

### LINKED ISSUE #25: Emoji Variation selectors

T0mstone:
This just changes all instances of `char` to `&str`/`String`.

Some notes:
- Allocating every variant in `build.rs` was the easiest way to do it. I'd have at least two other ideas about how to reduce allocations, but both require a bit more code complexity, so idk if it'd be worth it.
- The way the multi-character syntax interacts with `U+XXXX` escapes is by just treating each `U+` as the start of one and always taking the maximal hex value following that. I think this is fine, since we don't have any symbols or variants whose values contain hex digits anyway.

Future work would be updating `sym.txt` and `emoji.txt` to resolve #25. For that, I'd like to extend the syntax with suffixes `!text` and `!emoji` that get converted to `U+FE0E` and `U+FE0F`, so that it's less cryptic. Not sure if either of those changes (the suffix parsing and the actual updating of the module files) should be done in this PR or in a follow-up. I'd like some opinions on that.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -34,28 +34,28 @@ enum Def<'a> {
 
 /// A symbol, either a leaf or with modifiers with optional deprecation.
 enum Symbol<'a> {
-    Single(char),
-    Multi(Vec<(ModifierSet<&'a str>, char, Option<&'a str>)>),
+    Single(String),
+    Multi(Vec<(ModifierSet<&'a str>, String, Option<&'a str>)>),
 }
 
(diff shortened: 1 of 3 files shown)


T0mstone:
Ok I changed it to `\u{...}`

PR #93, opened 2025-06-30
Add bowties and deprecate `join`

T0mstone:
This is a necessary preliminary step to unblock #26, see also https://github.com/typst/codex/pull/26#issuecomment-2573429569.

I went with slightly different names from what's in the [Proposals Document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT) draft since `.l` and `.r` already have a use for the big operator, so using them entirely for filling felt wrong. I think the way they're called here is a better alternative.

The big ones don't have filled variants afaik, so I don't think defining them with a `.stroked` modifier as well makes much sense. Thus, it's now a trichotomy for this symbol: stroked, filled, or big.

As for the deprecation: I'm thinking of the process we discussed some time ago for stuff like this: Deprecate the original and add the new name in one typst version, then undeprecate and change the original name to the new symbol in the next typst version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,39 @@
 ## Unreleased
 
 ### New in `sym`
-- Miscellaneous Technical
+
+- Mathematical symbols
+  - `union.serif`: ∪︀
+  - `union.sq.serif`: ⊔︀
+  - `inter.serif`: ∩︀
+  - `inter.sq.serif`: ⊓︀
+  - `gt.double.nested`: ⪢
+  - `lt.double.nested`: ⪡
+  - `gt.arc` ⪧
+  - `gt.arc.eq` ⪩
+  - `lt.arc`: ⪦
+  - `lt.arc.eq`: ⪨
+  - `subset.approx`: ⫉
+  - `subset.closed`: ⫏
+  - `subset.closed.eq`: ⫑
+  - `subset.eq.dot`: ⫃
+  - `subset.equiv`: ⫅
+  - `subset.nequiv`: ⫋
+  - `subset.plus`: ⪿
+  - `subset.tilde`: ⫇
+  - `subset.times`: ⫁
+  - `supset.approx`: ⫊
+  - `supset.closed`: ⫐
+  - `supset.closed.eq`: ⫒
+  - `supset.eq.dot`: ⫄
+  - `supset.equiv`: ⫆
+  - `supset.nequiv`: ⫌
+  - `supset.plus`: ⫀
+  - `supset.tilde`: ⫈
+  - `supset.times`: ⫂
+  - `pee`: ℘
+
+- Miscellaneous technical
   - `bowtie.stroked`: ⋈
   - `bowtie.stroked.big`: ⨝
   - `bowtie.stroked.big.l`: ⟕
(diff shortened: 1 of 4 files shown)


MDLC01:
Ugh this is unfortunate. There is no way of deprecating the mai' variant of a symbol. What you did will show a warning for all variants of `join` as well...

Enivex:
It's somewhat unfortunate that l and r are used for entirely different things within the same symbol. But I guess it's unavoidable.

src/modules/sym.txt
```diff
@@ -584,10 +584,20 @@ diameter ⌀
 interleave ⫴
   .big ⫼
   .struck ⫵
+@deprecated: `join` is deprecated, use `bowtie.big` instead
 join ⨝
   .r ⟖
   .l ⟕
   .l.r ⟗
+bowtie
+  .stroked ⋈
```
Enivex, line 593: I think it makes more sense to just put this at the top level (without stroked), unless there are additional variants not included here.

T0mstone:
> It's somewhat unfortunate that l and r are used for entirely different things within the same symbol. But I guess it's unavoidable.

I think it's unambiguous enough as long as we don't have `bowtie.l`/`bowtie.r` on their own. (Tho these would currently automatically exist as shorthands, so #51 could improve the situation here).

src/modules/sym.txt
MDLC01: If we have `.stroked` we should have it for everything that is not filled. I agree having it is noisy, but it is better for consistency. The real problem is that now this is how it will be listed in the doc, so most users may not realize that `bowtie` works as well.

I would say we should keep `.stroked`, and improve the documentation's symbol lists as a separate project (e.g., by implementing https://github.com/typst/codex/issues/51).

A first step might be to announce stroked variants without the `.stroked` modifier in the changelog.

Enivex: Notice how the unicode names mention "black" specifically, but not "white" for the regular ones.

T0mstone: I really don't want the added complexity of enshrining a "canonical version" for some symbols by means of omitting certain modifiers for them. That's another data point users will have to remember and seems like a totally unnecessary restriction. It's conceptually so much simpler to just have both modifiers for everything with those two styles, whether one of them is canonical or not.

Enivex: I would say that having .stroked is `more` complexity, not less. Why should I have to arbitrarily have to remember which operators get the `stroked`, only because they have a `filled` variant? I do not like relying on fallback, because it's unpredictable and brittle to codex changes. Almost everyone who want these symbols will want the `.stroked` variants.

knuesel:
Compared to the old `join.l` we lose some clarity with `bowtie.big.l`: the name has nothing to suggest what happens on the left side. An idea could be to replace the `big` modifier with `join`:
```
bowtie
  .stroked ⋈
  .filled ⧓
  .filled.l ⧑
  .filled.r ⧒
  .join ⨝
  .join.r ⟖
  .join.l ⟕
  .join.l.r ⟗
```

src/modules/sym.txt
```diff
@@ -584,10 +584,20 @@ diameter ⌀
...
+  .filled ⧓
+  .filled.l ⧑
+  .filled.r ⧒
+  .big ⨝
```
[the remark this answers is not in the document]
MDLC01: Oh this is what I meant by "If we have `.stroked` we should have it for everything that is not filled." in https://github.com/typst/codex/pull/93#discussion_r2178385215. Sorry that that was unclear.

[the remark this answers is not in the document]
Enivex: I seem to just have missed that entirely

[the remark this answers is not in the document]
T0mstone: Done.

T0mstone:
I've added `.stroked` to the big variants as you suggested.

I've also changed the order so that `bowtie.l` and `bowtie.r` resolve to the large variants since that's probably more likely what the user would want (tho ideally with #51 those would just be forbidden here).

Enivex:
> I've added `.stroked` to the big variants as you suggested.
> 
> I've also changed the order so that `bowtie.l` and `bowtie.r` resolve to the large variants since that's probably more likely what the user would want (tho ideally with #51 those would just be forbidden here).

You also need to change the deprecation message

T0mstone:
> You also need to change the deprecation message

Not changing that was actually intentional. My thinking was that suggesting the full version might lead to worse UX since users may not know that the `.stroked` part is optional.

MDLC01:
I think we should suggest the simplest version of the name that we know we won't break in the near future. In this case, this would be without `.stroked`.

Fallback is a feature. It has many drawbacks, but there are also some cases where it makes sense to use it. I would argue this is on of them.

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+  - `bowtie.stroked.big.r`: ⟖
+  - `bowtie.stroked.big.l.r`: ⟗
+  - `bowtie.filled`: ⧓
+  - `bowtie.filled.l`: ⧑
+  - `bowtie.filled.r`: ⧒
+
+### Deprecated
+- `join` -> `bowtie.big`
+  - `join.l` and `join.r` to `bowtie.big.l` and `bowtie.big.r`, respectively
+  - `join.l.r` to `bowtie.big.l.r`
```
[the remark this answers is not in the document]
T0mstone: After trying a few things, nothing feels right here, so I'll just go with your suggestion. Maybe we should retcon the changelog later, rewriting all past entries to a consistent format...

MDLC01:
Remember that because this PR contains a breaking change, it will require a third approval.

PR #96, opened 2025-07-03
Fix docs

T0mstone:
These were slightly out of date after #92.

I've also taken the liberty to rephrase them a bit; Ever since #46, the crate-level docs have explained what a variant is, so we can just use that word in the functions' docs.

(Not sure whether this counts as meta or not)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,7 @@ pub enum Symbol {
 }
 
 impl Symbol {
-    /// Get the symbol's character for a given set of modifiers, alongside an optional deprecation
+    /// Get the symbol's variant for a given set of modifiers, alongside an optional deprecation
     /// message.
     pub fn get(&self, modifs: ModifierSet<&str>) -> Option<(&'static str, Option<&str>)> {
         match self {
@@ -79,7 +79,7 @@ impl Symbol {
         }
     }
 
-    /// The characters that are covered by this symbol.
+    /// Iterate over the variants of this symbol.
     ///
     /// Each variant is represented by a tuple `(modifiers, character, deprecation)`.
     pub fn variants(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,7 @@ impl Symbol {
 
     /// Iterate over the variants of this symbol.
     ///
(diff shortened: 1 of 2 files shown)


MDLC01:
I'd say this is meta but doesn't require Laurenz's approval since this is a minor documentation change.

MDLC01 (approved):

src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub enum Symbol {
 }
 
 impl Symbol {
-    /// Get the symbol's character for a given set of modifiers, alongside an optional deprecation
+    /// Get the symbol's variant for a given set of modifiers, alongside an optional deprecation
```
Enivex, line 71: Should this be string?

src/lib.rs
```diff
@@ -79,7 +79,7 @@ impl Symbol {
         }
     }
 
-    /// The characters that are covered by this symbol.
+    /// Iterate over the variants of this symbol.
     ///
     /// Each variant is represented by a tuple `(modifiers, character, deprecation)`.
```
Enivex, line 84: Should character be string?

T0mstone:
Found another small oversight that I've just lumped in this PR. Technically not documentation, but an assertion error message is close enough imo.

mkorje (approved):

src/lib.rs
```diff
@@ -2,10 +2,10 @@
 //!
 //! ## Model
 //! A [`Symbol`] is a collection of one or more _variants_. Each variant is
-//! identified by a set of [_modifiers_](ModifierSet) and has a single character
-//! as its value. The modifiers themselves can in principle be any non-empty
-//! strings that don't contain the character `.`, but codex only defines ones
-//! that are entirely made of ASCII alphabetical characters.
+//! identified by a set of [_modifiers_](ModifierSet) and has a string as its
```
mkorje, line 5: Actually, do we want to say here instead that it has a single grapheme cluster?

T0mstone: If at all, that should be another PR.

mkorje: Alright

mkorje (approved):

PR #97, opened 2025-07-03
Additional circled mathematical symbols

Enivex:
To my knowledge, these are the remaining circled mathematical operators in unicode.

There are a multitude more symbols in other categories, plus a generic composing circle (https://www.compart.com/en/unicode/U+20DD), but it seems natural to leave those for the future.

- Curiously, while there is a circled slanted division sign, there is no non-circled one.
- I debated whether I should add `.half` to `plus.o.l` etc, but ended up not doing that. The choice here is the same as in @MDLC01 's document.
- I think `times.o.o` would be a better name than `times.o.double`, but this is impossible right now.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -281,7 +281,7 @@ minus −
   .triangle ⨺
 div ÷
   .o ⨸
-  .o.slant ⦼
+  .slanted.o ⦼
   @deprecated: `div.circle` is deprecated, use `div.o` instead
   .circle ⨸
 times ×
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -252,6 +252,8 @@ prime ′
 // Arithmetic.
 plus +
   .o ⊕
+  .o.l ⨭
+  .o.r ⨮
   .o.arrow ⟴
   .o.big ⨁
   @deprecated: `plus.circle` is deprecated, use `plus.o` instead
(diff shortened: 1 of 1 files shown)


src/modules/sym.txt
```diff
@@ -279,11 +281,16 @@ minus −
   .triangle ⨺
 div ÷
   .o ⨸
+  .slanted.o ⦼
   @deprecated: `div.circle` is deprecated, use `div.o` instead
   .circle ⨸
 times ×
   .big ⨉
   .o ⊗
+  .o.double ⨷
```
MDLC01, line 290:
```suggestion
  .oo ⨷
```

I don't think this is better as `.o.double`, but this is also an option.

I personally don't like it because the meaning is not clear from the name, and it is reminiscent of `sym.oo`, which is completely different.

MDLC01 (approved):

src/modules/sym.txt
mkorje: I also think `.o.double` is better, and `.o.o` even better if/when that becomes possible.

mkorje (approved):

MDLC01 (changes requested):
> I think `times.o.o` would be a better name than `times.o.double`, but this is impossible right now.

How about we forget about this symbol for now, and come back to it when we have a clearer view of whether and how we want to allow duplicate modifiers?

That way, this PR can be merged and we don't keep yet another blocked PR.

MDLC01:
I feel like it's better not to introduce a minor feature if there is reasonable expectation that it will break at a later point.

Enivex:
Fair enough

PR #100, opened 2025-07-03
Add corner delimiters

Enivex:
In the previous large PR there was a suggestion to use `lt`, `lb`, `rt` and `rb` instead. However, I think it's important to signal which ones are opening (`.l`) and closing (`.r`) delimiters.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -50,6 +50,12 @@ shell
   .r.double ⟭
   .t ⏠
   .b ⏡
+bag
+  .l ⟅
+  .r ⟆
+mustache
+  .l ⎰
+  .r ⎱
 bar
   .v |
   .v.double ‖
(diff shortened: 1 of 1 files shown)


mkorje (approved):

MDLC01:
Note that there are other similar corners in Unicode: U+2E22 ⸢ TOP LEFT HALF BRACKET, U+2E23, U+2E24, and U+2E35. They are part of Supplemental Punctuation, while the codepoints added in this PR are from Miscellaneous Technical.

Enivex:
> Note that there are other similar corners in Unicode: U+2E22 ⸢ TOP LEFT HALF BRACKET, U+2E23, U+2E24, and U+2E35. They are part of Supplemental Punctuation, while the codepoints added in this PR are from Miscellaneous Technical.

Note that those are not delimiters, and seemingly not even intended for math. Not that this can't mean we can't add them in the future.

MDLC01 (approved):

mkorje (approved):

PR #104, opened 2025-07-04
Fix some gendered emoji

T0mstone:
### People with Bunny ears
These have been changed to be gender-neutral (in Unicode 9.0, according to [this](https://www.unicode.org/L2/L2024/24252-toned-multi-person-zwj-seq.pdf)).

I also added the gender-specific ZWJ sequences while I was at it, since multi-character symbols are now possible.

Source: https://www.unicode.org/reports/tr51/#multiperson_gender

### People holding hands
This is a ZWJ sequence, unlike the other three legacy variants.
I feel like this one only makes sense to also have, now that we can.

---

Not sure if we should actually merge this before [typst#6489](https://github.com/typst/typst/pull/6489), but it's probably fine since I'm pretty sure that that will land before the next release.

### DIFF
--- a/src/modules/emoji.txt
+++ b/src/modules/emoji.txt
@@ -377,7 +377,9 @@ cyclone 🌀
 dancing
   .man 🕺
   .woman 💃
-  .women.bunny 👯
+  .bunny 👯
+  .bunny.men 👯‍♂
+  .bunny.women 👯‍♀
 darts 🎯
 dash
   .wave.double 〰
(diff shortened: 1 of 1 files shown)


T0mstone:
For sure, but I came across these while implementing #51 and fixing them like this should be fine for now.

knuesel (approved):
This looks like a strict improvement. If we later do something smarter for gendered emojis in a breaking way, that would also break the existing `dancing.women.bunnny` anyway...

PR #105, opened 2025-07-04
Implement optional modifiers

### LINKED ISSUE #51: Improve forward compatibility with a notion of minimal modifier set

T0mstone:
Closes #51.

This is mostly done, but there are some unresolved questions and some issues and it still needs some finishing touches like documentation, which is why it's a draft for now.

This is also a breaking change since it changes the way `ModifierSet` works, so I should probably remember to bump the version in `Cargo.toml` before we merge this.

# Changes to observable syntax
The way I've implemented it, the `modifier?` syntax applies directly in the `ModifierSet`, so it would be exposed to the users in typst too. (Or typst would have to do some extra sanitizing).

And also, it's not really clear how this whole new resolution algorithm would interact with user-defined symbols from typst: I've kept the code in `best_match_in` mostly the same for now, but we could strip it down way more if the match was actually guaranteed to be unique. (But of course, this may not be true for user-defined symbols.)

# Changes to variant resolution
I have another branch where I wrote some ad-hoc code (mostly identical to what is now the `no_overlap` test) to brute-force check every possible modifier set for every symbol and generate a list of which ones differ between the old algorithm and the new one.

This is that list:
<details>
<pre><code>emoji.arrow : Some(("↙", None)) => Some(("➡", None))
emoji.bubble : Some(("💭", None)) => Some(("💬", None))
emoji.cloud.hidden : Some(("🌥", None)) => None
emoji.dancing.bunny : Some(("👯", None)) => None
emoji.dancing.women : Some(("👯", None)) => None
emoji.face.not : Some(("🫢", None)) => None
emoji.face.slight : Some(("🙁", None)) => None
emoji.face.withheld : Some(("🥹", None)) => None
emoji.faith : Some(("✝", None)) => None
emoji.faith.dot : Some(("🔯", None)) => None
emoji.finger.alt : Some(("☝", None)) => None
emoji.globe.af : Some(("🌍", None)) => None
emoji.globe.as : Some(("🌏", None)) => None
emoji.globe.au : Some(("🌏", None)) => None
emoji.globe.eu : Some(("🌍", None)) => None
emoji.handholding : Some(("👬", None)) => None
emoji.leaf.four : Some(("🍀", None)) => None
emoji.leaf.three : Some(("☘", None)) => None
emoji.monkey.not : Some(("🙉", None)) => None
emoji.moon.one : Some(("🌖", None)) => None
emoji.moon.face.three : Some(("🌜", None)) => None
emoji.moon.face.two : Some(("🌛", None)) => None
emoji.playback.once : Some(("🔂", None)) => None
emoji.playback.v : Some(("🔃", None)) => None
emoji.suit : Some(("♣", None)) => None
sym.angle.t : Some(("⦡", None)) => None
sym.angle.top : Some(("⦡", Some("`angle.spheric.top` is deprecated, use `angle.spheric.t` instead"))) => None
sym.arrow.half : Some(("↶", None)) => None
sym.arrows.stop : Some(("↹", None)) => None
sym.ballot.heavy : Some(("🗹", None)) => None
sym.colon.op : Some(("⫶", None)) => None
sym.dash.double : Some(("〰", None)) => None
sym.dash.three : Some(("⸻", None)) => None
sym.dash.two : Some(("⸺", None)) => None
sym.divides.rev : Some(("⫮", None)) => None
sym.dot.big : Some(("⨀", None)) => None
sym.emptyset.l : Some(("⦴", None)) => None
sym.emptyset.r : Some(("⦳", None)) => None
sym.eq.down : Some(("≒", None)) => None
sym.eq.up : Some(("≓", None)) => None
sym.gender.male.r : Some(("⚩", None)) => None
sym.gender.male.t : Some(("⚨", None)) => None
sym.gt.nested : Some(("⫸", None)) => None
sym.gt.slant : Some(("⩾", None)) => None
sym.integral.hook : Some(("⨗", None)) => None
sym.lt.nested : Some(("⫷", None)) => None
sym.lt.slant : Some(("⩽", None)) => None
sym.note.alt : Some(("♩", None)) => None
sym.note.beamed : Some(("♫", None)) => None
sym.note.slash : Some(("𝆔", None)) => None
sym.nothing.l : Some(("⦴", None)) => None
sym.nothing.r : Some(("⦳", None)) => None
sym.parallel.slanted : Some(("⧣", None)) => None
sym.parallel.slanted.tilde : Some(("⧤", None)) => None
sym.plus.arrow : Some(("⟴", None)) => None
sym.plus.big : Some(("⨁", None)) => None
sym.prec.curly : Some(("≼", None)) => None
sym.prec.curly.not : Some(("⋠", None)) => None
sym.prec.eq.not : Some(("⋠", None)) => None
sym.rest.measure : Some(("𝄩", None)) => None
sym.space.narrow : Some(("\u{202f}", None)) => None
sym.subset.not.sq : Some(("⋢", None)) => None
sym.succ.curly : Some(("≽", None)) => None
sym.succ.curly.not : Some(("⋡", None)) => None
sym.succ.eq.not : Some(("⋡", None)) => None
sym.suit : Some(("♣", None)) => None
sym.suit.filled : Some(("♣", None)) => None
sym.suit.stroked : Some(("♧", None)) => None
sym.supset.not.sq : Some(("⋣", None)) => None</code></pre>
</details>
Most of these should be net positives, but there are also some cases like <code>emoji.globe</code> where a variant has two optional modifiers, of which at least one needs to be present since it is not the default variant. This could in theory be encoded by duplicating the variant, but that'd be observable in the repr and also a bit hacky, so I haven't done it for now.

# Other stuff
I have some other minor things I'd like to say, but I'm too tired now, so I'll add them later.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,7 +107,7 @@ impl Symbol {
     /// Possible modifiers for this symbol.
     pub fn modifiers(&self) -> impl Iterator<Item = &str> + '_ {
         self.variants()
-            .flat_map(|(m, _, _)| m.into_iter())
+            .flat_map(|(m, _, _)| m.into_iter().map(|m| m.strip_suffix('?').unwrap_or(m)))
             .collect::<std::collections::BTreeSet<_>>()
             .into_iter()
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,4 +190,89 @@ mod test {
             assert_eq!(variants, control);
         }
     }
+
+    #[test]
+    fn no_overlap() {
+        recur("", ROOT);
+
+        /// Iterate over all symbols in a module, recursing into submodules.
+        fn recur(prefix: &str, m: Module) {
+            for (name, b) in m.iter() {
+                match b.def {
+                    Def::Module(m) => {
+                        let new_prefix = if prefix.is_empty() {
+                            name.to_string()
+                        } else {
+                            prefix.to_string() + "." + name
+                        };
+                        recur(&new_prefix, m);
+                    }
+                    Def::Symbol(s) => check_symbol(prefix, name, s),
+                }
+            }
+        }
+
+        /// Check the no overlap rule for a single symbol
+        fn check_symbol(prefix: &str, name: &str, sym: Symbol) {
+            // maximum number of modifiers per variant (we don't need to check more than this).
+            let max_modifs =
+                sym.variants().map(|(m, ..)| m.iter().count()).max().unwrap();
+            let modifs = sym.modifiers().collect::<Vec<_>>();
+            let max_index = modifs.len().saturating_sub(1);
+
+            for k in 0..=max_modifs {
+                let mut indices = (0..k).collect::<Vec<_>>();
+                loop {
+                    let mset = indices.iter().map(|i| modifs[*i]).fold(
+                        ModifierSet::<String>::default(),
+                        |mut res, m| {
+                            res.insert_raw(m);
+                            res
+                        },
+                    );
+
+                    if sym.variants().filter(|(m, ..)| mset.is_candidate(*m)).count() > 1
+                    {
+                        panic!(
+                            "Overlap in symbol {prefix}.{name} for modifiers {}",
+                            mset.as_str()
+                        );
+                    }
+
+                    if next_subseq(&mut indices, max_index).is_none() {
+                        break;
+                    }
+                }
+            }
+        }
+
+        /// Produce the next term in a sequence like
+        /// ```text
+        /// [0,1,2], [0,1,3], [0,1,4], [0,2,3], [0,2,4], [0,3,4], [1,2,3], [1,2,4], [1,3,4], [2,3,4]
+        /// ```
+        fn next_subseq(indices: &mut [usize], max_index: usize) -> Option<()> {
+            // invariant: indices.len() <= max_index + 1
+            match indices {
+                [] => None,
+                [single] => {
+                    if *single < max_index {
+                        *single += 1;
+                        Some(())
+                    } else {
+                        None
+                    }
+                }
+                [left @ .., last] => {
+                    assert_ne!(max_index, 0);
+                    if *last < max_index {
+                        *last += 1;
+                    } else {
+                        next_subseq(left, max_index - 1)?;
+                        *last = left.last().copied().map_or(*last, |x| x + 1);
+                    }
+                    Some(())
+                }
+            }
+        }
+    }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,7 +107,7 @@ impl Symbol {
     /// Possible modifiers for this symbol.
     pub fn modifiers(&self) -> impl Iterator<Item = &str> + '_ {
         self.variants()
-            .flat_map(|(m, _, _)| m.into_iter().map(|m| m.strip_suffix('?').unwrap_or(m)))
+            .flat_map(|(m, _, _)| m.into_iter().map(|m| m.name()))
             .collect::<std::collections::BTreeSet<_>>()
             .into_iter()
     }
@@ -170,7 +170,9 @@ mod test {
             };
             let variants = s
                 .variants()
-                .map(|(m, v, _)| (m.into_iter().collect::<BTreeSet<_>>(), v))
+                .map(|(m, v, _)| {
+                    (m.into_iter().map(|m| m.as_str()).collect::<BTreeSet<_>>(), v)
+                })
                 .collect::<BTreeSet<_>>();
             let control = control
                 .iter()
@@ -178,6 +180,7 @@ mod test {
                     (
                         ModifierSet::from_raw_dotted(m)
                             .into_iter()
+                            .map(|m| m.as_str())
                             .collect::<BTreeSet<_>>(),
                         v,
                     )
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -108,9 +108,7 @@ impl<S: Deref<Target = str>> ModifierSet<S> {
         let mut best_score = None;
 
         // Find the best table entry with this name.
-        for candidate in variants.filter(|(set, _)| {
-            set.required_is_subset(self.as_deref()) && self.is_subset(*set)
-        }) {
+        for candidate in variants.filter(|(set, _)| self.is_candidate(*set)) {
             let mut matching = 0;
             let mut total = 0;
             for modifier in candidate.0.iter() {
@@ -139,7 +137,13 @@ impl<S: Deref<Target = str>> ModifierSet<S> {
     /// Whether all *non-optional* modifiers in `self` are also present in `other`,
     /// optional or not.
     pub fn required_is_subset(&self, other: ModifierSet<&str>) -> bool {
-        self.iter().filter(|m| !m.is_optional()).all(|m| other.contains(m.as_str()))
+        self.iter()
+            .filter(|m| !m.is_optional())
+            .all(|m| other.contains(m.as_str()))
+    }
+
+    pub(crate) fn is_candidate(&self, other: ModifierSet<&str>) -> bool {
+        other.required_is_subset(self.as_deref()) && self.is_subset(other)
     }
 }
 
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -1,6 +1,6 @@
 use std::ops::Deref;
 
-/// A set of modifiers.
+/// A set of modifiers, some of which are marked as "optional".
 ///
 /// Beware: The [`Eq`] and [`Hash`] implementations are dependent on the
 /// ordering of the modifiers, in opposition to what a set would usually
@@ -15,16 +15,17 @@ pub struct ModifierSet<S>(
 
 impl<S: Deref<Target = str>> ModifierSet<S> {
     /// Constructs a modifier set from a string, where modifiers are separated
-    /// by the character `.`.
+    /// by the character `.` and optional modifiers end in the character `?`.
+    /// (The `?` is not considered part of the modifier.)
     ///
     /// `s` should not contain any empty modifiers (i.e. it shouldn't contain
-    /// the sequence `..`) and no modifier should occur twice. Otherwise,
+    /// the sequences `..` or `.?.`) and no modifier should occur twice. Otherwise,
     /// unexpected errors can occur.
     pub fn from_raw_dotted(s: S) -> Self {
         // Checking the other requirement too feels like it would be a bit too
         // expensive, even for debug mode.
         debug_assert!(
-            !s.contains(".."),
+            !s.contains("..") && !s.contains(".?."),
             "ModifierSet::from_dotted called with string containing empty modifier"
         );
         Self(s)
@@ -33,6 +34,11 @@ impl<S: Deref<Target = str>> ModifierSet<S> {
     /// Whether `self` is empty.
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
+    }
+
+    /// Whether `self` has any optional modifiers.
+    pub fn has_optional(&self) -> bool {
+        self.0.ends_with('?') || self.0.contains("?.")
     }
 
     /// Gets the string of modifiers separated by `.`.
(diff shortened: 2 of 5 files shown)


T0mstone:
> Maybe a solution would be to define
> 
> ```
> globe
>   .as?.au 🌏
>   .as.au? 🌏
>   .eu?.af 🌍
>   .eu.af? 🌍
> ```
> 
> and have the validation code recognize that `.as?.au` and `.as.au?` don't clash because it's the same variant?

Yes, that's what I meant with "duplicating the variant". Tho your example wouldn't work since `globe.as.au` would be ambiguous, but the following would work:
```
globe
  .as.au? 🌏
  .au 🌏
  .eu.af? 🌍
  .af 🌍
```
The issue with this is something that ties into the next thing I want to talk about here (see my comment below this).

T0mstone:
These changes to `ModifierSet` require us to add some extra code to typst if we don't want them to be observable to the end users, because if we use it as-is, the `modifier?` syntax will be exposed. (I've already said this above.)

Now my point here is that I actually think exposing this is a good thing and we should give users access to the new resolution system too, so custom symbols aren't second-class citizens.
But that poses the question of what to do when it comes to verifying unique abbreviations.

In codex, we can just run the test `no_overlap`, which is very inefficient and takes a few seconds on my machine, but typst has stricter performance requirements.
In particular, I see two options there:
1. Somehow implement uniqueness checking for typst as well.(maybe pull it from the test into a global function that works with any `Symbol`-likes) This would probably require some heavy optimization of which I'm not sure whether it's possible.
2. Let users define symbols with overlapping abbreviations and retain the extra code in the loop in `best_match_in` to handle these ambiguities when they exist. This has a huge performance benefit, but comes at the cost of not letting package authors get the same guarantees that codex now has for forward-compatibility. (Maybe we could expose a uniqueness check explicitly as a method on typst's `symbol`, so package authors can incorporate it into their tests?)

T0mstone:
Another small thing that confused me a bit (and why I had to double-check that `sym.supset.not.sq` should be invalid) is that DejaVu Sans Mono displays `⋣` without the "or equal" part. I guess it's technically not totally wrong since that's also a notation some set theorists use, but it really feels like an error.

knuesel:
> Somehow implement uniqueness checking for typst as well.(maybe pull it from the test into a global function that works with any Symbol-likes) This would probably require some heavy optimization of which I'm not sure whether it's possible.

I think a fast implementation in Typst should be possible. To detect overlaps it should be sufficient to populate a dictionary where the keys are all "aliases" of all variants. The initial list of keys could be generated statically when building the Typst executable. Then checking overlap for a new symbol would just be a dict insertion (negligible runtime) which would fail if the key already exists (unless I'm missing something and the check must be more sophisticated?).

src/shared.rs
```diff
@@ -123,9 +158,49 @@ impl<S: Default> Default for ModifierSet<S> {
...
+    pub fn as_str(self) -> &'a str {
+        self.0
+    }
+
+    pub fn name_and_is_optional(self) -> (&'a str, bool) {
+        match self.0.strip_suffix('?') {
+            Some(name) => (name, true),
+            None => (self.0, false),
+        }
+    }
```
MDLC01, line 174: As far as I can tell, this function isn't used currently. Given that it is a combination of the two functions below, we can probably remove this one.

src/lib.rs
```diff
@@ -170,14 +170,17 @@ mod test {
             };
             let variants = s
                 .variants()
-                .map(|(m, v, _)| (m.into_iter().collect::<BTreeSet<_>>(), v))
+                .map(|(m, v, _)| {
+                    (m.into_iter().map(|m| m.as_str()).collect::<BTreeSet<_>>(), v)
```
MDLC01, line 174:
```suggestion
                .map(|(ms, v, _)| {
                    (ms.into_iter().map(|m| m.as_str()).collect::<BTreeSet<_>>(), v)
```

Or something else to make it clear that this is a modifier _set_.

src/lib.rs
```diff
@@ -187,4 +190,89 @@ mod test {
...
+                }
+            }
+        }
+
+        /// Produce the next term in a sequence like
+        /// ```text
+        /// [0,1,2], [0,1,3], [0,1,4], [0,2,3], [0,2,4], [0,3,4], [1,2,3], [1,2,4], [1,3,4], [2,3,4]
+        /// ```
+        fn next_subseq(indices: &mut [usize], max_index: usize) -> Option<()> {
+            // invariant: indices.len() <= max_index + 1
```
MDLC01, line 254: This should be specified in the documentation for this function.

src/lib.rs
```diff
@@ -187,4 +190,89 @@ mod test {
...
+                }
+                [left @ .., last] => {
+                    assert_ne!(max_index, 0);
+                    if *last < max_index {
+                        *last += 1;
+                    } else {
+                        next_subseq(left, max_index - 1)?;
+                        *last = left.last().copied().map_or(*last, |x| x + 1);
+                    }
+                    Some(())
```
MDLC01, line 273:
```suggestion
                    if *last < max_index {
                        *last += 1;
                    } else {
                        *last = next_subseq(left, max_index - 1)?;
                    }
                    Some(*last)
```

This seems more idiomatic to me.

src/lib.rs
```diff
@@ -187,4 +190,89 @@ mod test {
...
+            for k in 0..=max_modifs {
+                let mut indices = (0..k).collect::<Vec<_>>();
+                loop {
+                    let mset = indices.iter().map(|i| modifs[*i]).fold(
+                        ModifierSet::<String>::default(),
+                        |mut res, m| {
+                            res.insert_raw(m);
+                            res
+                        },
+                    );
```
MDLC01, line 232: Maybe we should have an `impl<S: Deref<Target = str>> FromIterator<S> for ModifierSet<String>` instead.

src/shared.rs
```diff
@@ -15,16 +15,17 @@ pub struct ModifierSet<S>(
 
 impl<S: Deref<Target = str>> ModifierSet<S> {
     /// Constructs a modifier set from a string, where modifiers are separated
-    /// by the character `.`.
+    /// by the character `.` and optional modifiers end in the character `?`.
+    /// (The `?` is not considered part of the modifier.)
     ///
     /// `s` should not contain any empty modifiers (i.e. it shouldn't contain
-    /// the sequence `..`) and no modifier should occur twice. Otherwise,
+    /// the sequences `..` or `.?.`) and no modifier should occur twice. Otherwise,
```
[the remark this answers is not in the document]
T0mstone: Good point. To add to that, the previous comment had already been wrong (not just the sequence `..`, also a `.` at the end could create an empty modifier).

src/shared.rs
T0mstone: Yeah, my thought was to allow users to avoid the double check, but I guess the compiler is probably smart enough to just optimize that out. Maybe the functions need `#[inline]` for that, so I'll add that as well (they're also small enough to warrant that anyway).

src/lib.rs
T0mstone: I really don't like that, see https://github.com/typst/codex/pull/46#issuecomment-2683218146.
Maybe we could have an inherent method `from_iter_raw` that documents all of those requirements (`S::default()` must be empty, each item must be a single modifier). To use that like `collect`, we'd also need an iterator extension trait, which seems overkill, but maybe calling it directly on an iterator is already ergonomic enough?

MDLC01: Oh indeed you are right! I had forgotten I already suggested a similar thing before.

I think having `from_iter_raw` could be useful.

src/shared.rs
```diff
@@ -36,9 +36,9 @@ impl<S: Deref<Target = str>> ModifierSet<S> {
         self.0.is_empty()
     }
 
-    /// Whether `self` has any optional modifiers.
+    /// Whether `self` has any optional modifier(s).
```
MDLC01, line 39: Singular "modifier" would be the most grammatically correct here I think, after "any".

src/shared.rs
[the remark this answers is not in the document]
MDLC01: Good catch!

src/lib.rs
```diff
@@ -264,11 +272,12 @@ mod test {
                 }
                 [left @ .., last] => {
                     assert_ne!(max_index, 0);
+                    assert_ne!(left.len(), 0);
```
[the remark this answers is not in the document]
T0mstone: The length must be at least 2 since lengths 0 and 1 are handled in the match arms above.

src/shared.rs
T0mstone: No, I don't think so; See for example https://ell.stackexchange.com/a/1177. Singular sounds wrong to me here, so I went with parenthesized plural for maximum clarity, since it can be one or more.

src/shared.rs
```diff
@@ -123,9 +158,44 @@ impl<S: Default> Default for ModifierSet<S> {
     }
 }
 
+#[derive(Copy, Clone)]
+pub struct Modifier<'a>(&'a str);
```
laurmaedje, line 162: This type appears in the public API, but is not re-exported, so it's a deadlink in the docs. If it's supposed to properly public, it would also be good to have a bit of docs on it.

laurmaedje (commented):
I didn't review the code too closely, but I'm okay with this in principle. I just noticed one thing while going through it. Otherwise, I'd defer to @MDLC01's judgement.

src/shared.rs
MDLC01: Ok let's go with "(s)" then

T0mstone:
This is a PR to codex, so it *cannot* decide what users can or can't do. That would have to happen in a follow-up PR to typst.

PR #106, opened 2025-07-04
Untangle fence/zig-zag mess

Enivex:
- ⦙ was misplaced under fence, presumably because its unicode name is "Dotted Fence". However, I believe this name simply comes from its default math class. It more correctly belongs under `dots`, even if their math classes are different. It's latex name is `\fourvdots`
- I renamed `fence` to `zigzag`, which is a more descriptive name that's also consistent with their latex naming. (Note that the look of these in New Computer Modern Math is very strange relative to other math fonts.)
- I added the missing ⦚ (class "fence") at the top level of zigzag.

Left to right: zigzag, zigzag.l, zigzag.r, zigzag.l.double, zigzag.r.double, dots.v.quad
Top to bottom: Noto Sans Math, Stix Two Math, New Computer Modern Math.

image


Enivex:
I'll write more later, but for now I'll just add this

https://www.unicode.org/L2/L2000/00094-n2191.pdf

It specifically specifies "four close dots vertical"

T0mstone (changes requested):

src/modules/sym.txt
```diff
@@ -60,11 +60,21 @@ bar
...
   .r ⧙
+  @deprecated: `fence.r.double` is deprecated, use `zigzag.r.double` instead
   .r.double ⧛
+  @deprecated: `fence.dotted` is deprecated, use `dots.v.quad` instead
   .dotted ⦙
+zigzag ⦚
+  .l ⧘
+  .l.double ⧚
+  .r ⧙
+  .r ⧛
```
```suggestion
  .r.double ⧛
```

MDLC01:
I just realized that U+2E3E ⸾ WIGGLY VERTICAL LINE is also a thing. It does not have a math class, but one may argue that this is what U+29D8, U+29D9, U+29DA, and U+29DB are variants of, and that U+299A ⦚ VERTICAL ZIGZAG LINE is its own separate thing.

PR #107, opened 2025-07-04
`angle.oblique` should be `angle.obtuse`

Enivex:
While oblique can apparently be used to refer to any angle that's not right, that's clearly not the intended meaning here. This is specifically an obtuse angle (that is, larger than 90 degrees).

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -77,6 +77,8 @@ angle ∠
   .acute ⦟
   .arc ∡
   .arc.rev ⦛
+  .obtuse ⦦
+  @deprecated: `angle.oblique` is deprecated, use `angle.obtuse` instead
   .oblique ⦦
   .rev ⦣
   .right ∟


knuesel:
A shorter alternative would be `.blunt` but I'm also fine with `.obtuse`.

T0mstone:
> that's clearly not the intended meaning here.

I sort of disagree. The Unicode name for this is "OBLIQUE ANGLE OPENING UP" and as usual with Unicode there's no further clarifying information other than how this glyph looks in some fonts.

After a lot of investigation, I managed to find out that the official Unicode tables (in this case https://www.unicode.org/charts/PDF/U2980.pdf) use STIX for their example glyphs. There's also some sort of comment on this in the Unicode archives: https://www.unicode.org/L2/L2013/13221r-comments-on-stix-fonts.pdf:
> “Math STIX font set is a better representation of the mathematical characters encoded in the standard.”

(quote by an individual)

So it might be fine™ to just roll with it and go by what they look like in STIX, but I'm still a bit unsure about this.

Enivex:
There are plenty of examples of Unicode names that are misleading or straight up incorrect. I'm on the go, so I'm unable to look up the proposal right now. In any case, the reality is that font authors universally depict this symbol as obtuse

dccsillag:
Also never heard of a blunt angle. Obtuse also sounds like a safer bat to me.

Enivex:
This one needs one reapproval

knuesel (approved):

PR #110, opened 2025-07-04
Add missing arc modifier to `angle.right.dot`, rename `angle.right.sq`

Enivex:
⦝ is specifically ⊾ with a dot, and `.sq` should mean a "squarified" version of a symbol (as opposed to `.square`)

Alternatively we could rename `angle.right.sq` to `angle.right.square` instead, but that would be a breaking change.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -83,7 +83,9 @@ angle ∠
   .right.rev ⯾
   .right.arc ⊾
   .right.arc.dot ⦝
-  .right.arc.sq ⦜
+  .right.square ⦜
+  @deprecated: `angle.right.sq` is deprecated, use `angle.right.square` instead
+  .right.sq ⦜
   .s ⦞
   .spatial ⟀
   .spheric ∢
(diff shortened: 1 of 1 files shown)


MDLC01:
I am fine with the change from `angle.right.dot` to `angle.right.arc.dot`.

I'm not sure why we would want ⦜ to be a variant of `angle.right.arc`?

MDLC01:
You can also see it as one having an arc and the other having a square. An arc is typically a portion of a circle.

I don't think formal definitions matter that much (what does is users' perception of what an "arc" is), but, to quote [Wikipdia](https://en.wikipedia.org/wiki/Curve#Differentiable_arc), "an **arc** [...] is a connected subset of a differentiable curve." A square is not a differentiable curve.

T0mstone:
If we're going to start making modifiers globally consistent, I'm going to demand keeping them in a markdown file in the repo (like I already mentioned once or twice on Discord)

Enivex:
> If we're going to start making modifiers globally consistent, I'm going to demand keeping them in a markdown file in the repo (like I already mentioned once or twice on Discord)

It's not always possible to make everything 100% consistent, but it's always been a goal to try. I guess we could try to write down some general guidelines at least.

Enivex:
> Then I would be in favor of using `.square`.

I've made the change.

T0mstone (approved):

PR #111, opened 2025-07-06
Consistently use `.slant`

Enivex:
Take two.

- Add the most closely related additional variants. There are a bunch more, but those can wait to avoid more bikeshedding.
- This should unblock https://github.com/typst/codex/pull/36
- We should aim to have a way for users to consistently use the slanted variants by default in the future, as these aren't very ergonomic.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -313,7 +313,9 @@ eq =
   .equi ≚
   .est ≙
   .gt ⋝
+  .gt.slant ⪖
   .lt ⋜
+  .lt.slant ⪕
   .m ≞
   .not ≠
   .prec ⋞
@@ -333,9 +335,11 @@ gt >
   .double ≫
   .eq ≥
   .eq.slant ⩾
+  .eq.slant.dot ⪀
   .eq.lt ⋛
   .eq.not ≱
   .equiv ≧
+  .equiv.slant ⫺
   .lt ≷
   .lt.not ≹
   .neq ⪈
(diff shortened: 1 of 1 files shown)


MDLC01:
Shouldn't we wait before we have a final solution to the `.slant` vs. `.alt` vs. `.seq` problem before adding all those?

Enivex:
> Shouldn't we wait before we have a final solution to the `.slant` vs. `.alt` vs. `.seq` problem before adding all those?

I answered this on Discord, but I think it's valuable to have it here. It seemed that getting consensus about the `seq` solution was pretty much impossible.

My thinking is that we go for this now, but try to find a more ergonomic way for those who prefer the slanted variants to use those by default in the future.

By the way: We should probably also discuss what to do about `parallel.slanted.*` and `div.slanted.*`. Those are slightly different, in that it's the symbol itself that is slanted, and not the modifier. Not sure that's a big deal though.

MDLC01:
For reference, this PR supersedes https://github.com/typst/codex/pull/89.

`.slant` vs. `.slanted` is indeed an inconsistency, but I agree we should merge this for now. The inconsistency already exists anyway, and this will allow us to move forward with #36 as you mentioned.

knuesel:
> There are a bunch more, but those can wait to avoid more bikeshedding.

Would you have a list at hand? I think this PR introduces good, consistent naming but it would be nice to have a sense of how it would fit with those other symbols.

knuesel (approved):

T0mstone (changes requested):

MDLC01:
Superseded by https://github.com/typst/codex/pull/155.

MDLC01:
To clarify, I closed this because the changelog needs to be updated and I cannot push to the PR myself.

PR #112, opened 2025-07-15
Add contributing guidelines

T0mstone:
As mentioned in https://github.com/typst/codex/pull/110#issuecomment-3038954763.

These conventions are clearly contributing guidelines, so I went with `CONTRIBUTING.md` for the file name. That way we can also add other contributing guidelines in the future, if we want to.

If you want to reword something or add other conventions that I may have missed, don't hesitate to leave a comment.

From my side, there's only one unresolved question left here (which I also left as a TODO comment in the file itself), namely whether we want to write something about when we choose e.g. `.t.r` over `.tr` for corners (cf. https://github.com/typst/codex/pull/100).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Modifiers are order-independent, so the latter can also be referred to as `arrow
 Additionally, not all modifiers have to be specified, in which case the best match[^match]
 will be taken. For example, `⇒` can also be referred to as `arrow.double`.
 Groups of related symbols are collected into *modules*. Modules can also contain other modules.
-Codex exports two top-level modules: `sym` for for text-style symbols and `emoji` for emoji;
+Codex exports two top-level modules: `sym` for text-style symbols and `emoji` for emoji;
 Their source code is found in `src/modules/`.
 
 If you need help with a contribution, you can also ask us [on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306).


CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+
+General conventions:
+- Modifiers are entirely lowercase.
+- Modifiers that correspond to symbols (by having the same name) have six established meanings:
+	- The symbol is added around the base like an accent or an attachment (smaller than the base), e.g. `eq.quest`.
+	- The symbol is stacked below the base, e.g. `gt.lt`.
+	- The symbol is stacked to the right of the base, e.g. `colon.eq`.
+	- The symbol is overlaid at the center of the base, e.g. `integral.dash`.
+	- The symbol surrounds the base, e.g. `plus.square`.
+	- The symbol is placed inside the base, e.g. `triangle.stroked.dot`.
```
MDLC01, line 14: I think this could be grouped with the first bullet point. In both cases, the idea is that the symbol visually appears as a subordinate addition to the base.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+Established generic modifiers:
+- `.l`/`.r`/`.t`/`.b`: The four main directions (left/right/top/bottom).
+	- For delimiters, `.l` means opening and `.r` means closing (see [#100](https://github.com/typst/codex/pull/100)).
```
MDLC01, line 17: Why link to #100 specifically? I think an example such as `paren.l` and `bracket.r` would be better.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.tl`/`.tr`/`.bl`/`.br`: The four corners
+	<!-- TODO: Do we have or want to have conventions about when to choose `.tl` vs. `.t.l`? -->
```
MDLC01, line 19: `.tl` designates the top-left corner, whereas `.t.l` (which AFAIK doesn't really exists at the moment appart from `bracket.{l,r}.tick.{t,b}`) is a combination of both `.t` and `.l`. For example, `arrow.t.r` would be an arrow pointing both top and to the right, similarly to how `arrow.l.r` is an arrow pointing both left and right.

The case of corner delimiters is special because we want to signal opening and closing delimiters, as explained in #100.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.double`, `.triple`, `.quad`: A symbol that has 2-4 of something.
+- `.stroked`/`.filled`: A symbol that has an empty/filled interior.
+	(They correspond to Unicode's "white"/"black".)
+- `.tiny`/`.small`/`.medium`/`.big`: A geometric shape with a certain size.
+- `.light`/`.heavy`: A shape with a certain stroke weight.
+- `.dotted`: A shape with a dotted line instead of a full stroke.
+Established concrete modifiers:
```
MDLC01, line 26: I'm not sure I understand the difference between this and "generic modifiers".

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.big`: A big (n-ary) version of an operator.
```
MDLC01, line 27:
```suggestion
- `.big`: A large version of an operator.
```

We might as well use [UTR #25](https://www.unicode.org/Public/math/latest/MathClassEx-15.html) terminology, as `.big`, for mathematical symbols, implies a change of math class to Large, typically from Binary.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.alt`: An alternative version that is still more or less the same symbol.
```
MDLC01, line 28:
```suggestion
- `.alt`: An alternate glyph for the symbol.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.not`: A symbol with a (typically diagonal) line through it.
```
MDLC01, line 29:
```suggestion
- `.not`: A negation of the symbol.
```

This is more than just a diagonal line. It's the semantic idea of negation, which happens to be almost always represented with a diagonal line through.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+- `.o`: A symbol with a circle around it (see [#62](https://github.com/typst/codex/pull/62))
+- `.sq`: A "squarified" version of a symbol (see [#110](https://github.com/typst/codex/pull/110))
+- `.rev`: A horizontally mirrored version of a symbol (see [#108](https://github.com/typst/codex/issues/108))
+- `.inv`: Either vertically mirrored or a 180° rotated version of a symbol (see [#108](https://github.com/typst/codex/issues/108))
+- `.cw`/`.ccw`: Clockwise/Counterclockwise
+Established exceptions:
+- When `.eq` is used in the second sense (stacked below), it only adds a single line and not two.
```
MDLC01, line 36: "the second sense" here refers to something that is too far away from here. This exception should probably be moved there instead.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+	For two lines below, `.equiv` is used.
+- `.not` (see above) does not correspond to the symbol `not` despite having the same name.
```
MDLC01, line 38: I don't think this should be an exception. Instead, I think the sentence

> Modifiers that correspond to symbols (by having the same name) have six established meanings:

should be rephrased to something like

> When a symbol is added to a base, the symbol name is used as a modifier on the base. This can have the following meanings:

MDLC01 (commented):
My personal view is that this document should be used as a way to quickly find precedents, but that new names should always be motivated by existing symbols. In other words, those guidelines should reflect the existing conventions and shouldn't be used as a source of truth--the only source of truth is the existing names. For this reason, I strongly believe every rule in this document should be supported by one or multiple examples. This is reflected in my proposed rephrasing of the first paragraph.

Unrelatedly, if this document aims to be contributor guidelines, it might benefit from some of the links from the "Useful links" section of the [proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT). Maybe even a link to the document itself.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+because of how messy Unicode can be, but you should adhere to them if possible.
+
+General conventions:
+- Modifiers are entirely lowercase.
+- Modifiers that correspond to symbols (by having the same name) have six established meanings:
+	- The symbol is added around the base like an accent or an attachment (smaller than the base), e.g. `eq.quest`.
+	- The symbol is stacked below the base, e.g. `gt.lt`.
+	- The symbol is stacked to the right of the base, e.g. `colon.eq`.
+	- The symbol is overlaid at the center of the base, e.g. `integral.dash`.
+	- The symbol surrounds the base, e.g. `plus.square`.
```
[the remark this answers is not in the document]
T0mstone: Maybe we should reopen the issue, given that this part still remains unaddressed. Solving this now might be tricky tho, because it would ideally be consistent with `.o`, but idk if there's a way to do that.

CONTRIBUTING.md
T0mstone: Fair enough. Should I add a clarification for this or do you think it just goes without saying?

CONTRIBUTING.md
T0mstone: Generic modifiers have a broader meaning that can be expressed in various ways (the directions are a clear example of this) while concrete ones have one explicit geometric transformation that they do.

I guess `.alt` should go into the generic ones then...🤔

T0mstone:
Another question that should be resolved: Should `.cw` and `.ccw` be generic modifiers instead?

Or, going further, perhaps the whole distinction should be removed entirely? (Tho I personally like it.)

T0mstone:
Changing the title to reflect the broader scope.

MDLC01 (commented):
> Or, going further, perhaps the whole distinction should be removed entirely?

Maybe just ordering the modifiers from most concrete to least concrete would be enough. I personally don't see a clear cut distinction.

CONTRIBUTING.md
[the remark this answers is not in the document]
MDLC01: I created https://github.com/typst/codex/issues/113 as a separate issue instead.

CONTRIBUTING.md
T0mstone: Personally I like it more the way it is now, because the generic ones theoretically have more uses, so they feel more important.

Enivex:
I think it would be helpful to have them organized alphabetically, at least within a category.

CONTRIBUTING.md
T0mstone: Is 10d25457d54a380f91b2bb16f0d9199fe37b30f8 good?

T0mstone:
> Another question that should be resolved: Should `.cw` and `.ccw` be generic modifiers instead?

I've now on "yes".

MDLC01 (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+existing ones. Below are some guidelines based on existing symbols. These aren't
+always hard rules, especially because of how messy Unicode can be, but you should
+adhere to them if possible.
+
+### General Conventions
+- English words use US spelling.
+- Modifier and module names are entirely lowercase.
+- Symbol names are lowercase unless the symbol is an uppercase letter.
+- Symbol names should be at least two characters long so they can be used easily in Typst's math mode.
+- When a symbol is added to a base, the symbol name is used as a modifier on the base.[^modifname]
```
knuesel:
```suggestion
- For variants that add a geometric shape to a base symbol, the shape name is used as modifier with one of the following meanings:[^modifname]
```

I think this clarifies it a bit (I was initially confused by the double meaning of "symbol" and no mention of variant).

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+Contributions are welcome! This document provides some resources and guidelines to help with the process.
+
+Symbol definitions are in `src/modules/`; The syntax should be mostly self-evident.
+If you need help with a contribution, you can ask us [on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306).
+
+## Proposals
+Proposals used to be written in the [Proposals document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT),
+although it is now preferred to have a GitHub issue for each one.
+Nonetheless, the document still contains a lot of useful information.
+
```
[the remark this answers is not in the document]
T0mstone: A bit of this is already in the `lib.rs` doc-comment, but I suppose it doesn't hurt repeating it here.

CONTRIBUTING.md
T0mstone: I slightly disagree. This is specifically about when we already have an *existing symbol* for the added shape. That's what I meant by "When a symbol is added to a base".

knuesel: I see. But don't we want to apply the same rules regardless of whether the shape is already its own symbol? For example I think we would want `plus.square` to follow these rules, even if we didn't have a `square` symbol on its own?

PR #114, opened 2025-07-16
Specify all presentation sequences

### LINKED ISSUE #21: Hooked arrow symbols should use a text variation selector (VS15) by default
### Description

Currently, `#sym.arrow.hook` and its left variant are rendered as emoji, which is inconsistent with the rest of the arrows in this pseudo-module (all text). For some odd reasons, [the other arrows that have an emoji variant](https://en.wikipedia.org/wiki/Arrows_(Unicode_block)#Emoji) (e.g. `#sym.arrow.l`) properly render as text.

Given that there is already an `#emoji.arrow.hook` to obtain this emoji, it would make more sense that `#sym.arrow.hook` forces text rendering.

image

This may be linked to typst/codex#22. _In case you're wondering (and for indexing purposes): I'm trying to implement backlinks (similar to footnote backlinks) and the ↪︎ symbol is what's used traditionally for that purpose._ 

### Reproduction URL

https://typst.app/project/rHcPSm8pdJ6xy3dxLmijAQ

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

### LINKED ISSUE #23: emoji.arrow.r.filled should include the variation selector
### Description

`#emoji.arrow.r.filled` is `\u{27A1}` but it should be the emoji sequence `\u{27A1}\u{FE0F}` (with the Variation Selector 16).

Most external emoji pickers (Windows, Gnome, copy and paste website) correctly use the [emoji sequence](https://unicode.org/emoji/charts/emoji-sequences.html).

Reported on [typst-svg-emoji#7](https://github.com/polazarus/typst-svg-emoji/issues/7).

Also, it seems the web app with the experimental emoji support enforces the use of Twemoji for the lone codepoint.

### Reproduction URL

_No response_

### Operating system

Windows, Linux, macOS

### Typst version

- [X] I am using the latest version of Typst

### LINKED ISSUE #25: Emoji Variation selectors

MDLC01:
This PR fixes #21, fixes #23, and closes #25, by adding the appropriate variation selectors to symbols that exist in both `sym` and `emoji`. I verified that all the variation sequences are defined in by Unicode.[^1][^2]

[^1]: https://www.unicode.org/reports/tr51/#Emoji_Variation_Sequences
[^2]: https://www.unicode.org/Public/UCD/latest/ucd/emoji/emoji-variation-sequences.txt

I have marked this PR as a draft because the next step is to add variation selectors to all symbols that allow it, whether present in both `sym` and `emoji` or not, to prevent ambiguity.

This made me realize that some emojis are poorly named, but improving this is a task for a separate PR.

Related: https://github.com/typst/typst/pull/6489#discussion_r2211811225.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,8 @@
+use self::shared::ModifierSet;
 use std::fmt::Write;
 use std::iter;
 use std::iter::Peekable;
 use std::path::Path;
-
-use self::shared::ModifierSet;
 
 type StrResult<T> = Result<T, String>;
 
@@ -192,23 +191,60 @@ fn validate_ident(string: &str) -> StrResult<()> {
     Err(format!("invalid identifier: {string:?}"))
 }
 
-/// Extracts the value of a variant, parsing `\u{XXXX}` escapes
-fn decode_value(text: &str) -> StrResult<String> {
-    let mut iter = text.split("\\u{");
-    let mut res = iter.next().unwrap().to_string();
-    for other in iter {
-        let (hex, rest) = other.split_once("}").ok_or_else(|| {
-            format!("unclosed unicode escape \\u{{{}", other.escape_debug())
-        })?;
-        res.push(
-            u32::from_str_radix(hex, 16)
-                .ok()
-                .and_then(|n| char::try_from(n).ok())
-                .ok_or_else(|| format!("invalid unicode escape \\u{{{hex}}}"))?,
-        );
-        res += rest;
+/// Extracts the value of a variant, parsing `\u{XXXX}` and other escapes.
+fn decode_value(mut text: &str) -> StrResult<String> {
+    let mut result = String::new();
+    loop {
+        if let Some(rest) = text.strip_prefix("\\u{") {
+            let Some((code, tail)) = rest.split_once('}') else {
+                return Err(format!(
+                    "unclosed Unicode escape: \\u{{{}",
+                    rest.escape_debug()
+                ));
+            };
+            result.push(
+                u32::from_str_radix(code, 16)
+                    .ok()
+                    .and_then(|n| char::try_from(n).ok())
+                    .ok_or_else(|| format!("invalid Unicode escape \\u{{{code}}}"))?,
+            );
+            text = tail;
+        } else if let Some(rest) = text.strip_prefix("\\vs{") {
+            let Some((value, tail)) = rest.split_once('}') else {
+                return Err(format!("unclosed VS escape: \\vs{{{}", rest.escape_debug()));
+            };
+            let vs = match value {
+                "1" => '\u{fe00}',
+                "2" => '\u{fe01}',
+                "3" => '\u{fe02}',
+                "4" => '\u{fe03}',
+                "5" => '\u{fe04}',
+                "6" => '\u{fe05}',
+                "7" => '\u{fe06}',
+                "8" => '\u{fe07}',
+                "9" => '\u{fe08}',
+                "10" => '\u{fe09}',
+                "11" => '\u{fe0a}',
+                "12" => '\u{fe0b}',
+                "13" => '\u{fe0c}',
+                "14" => '\u{fe0d}',
+                "15" | "text" => '\u{fe0e}',
+                "16" | "emoji" => '\u{fe0f}',
+                code => return Err(format!("invalid VS escape: \\vs{{{code}}}")),
+            };
+            result.push(vs);
+            text = tail;
+        } else if let Some((prefix, tail)) = text.find('\\').map(|i| text.split_at(i)) {
+            if prefix.is_empty() {
+                return Err(format!("invalid escape sequence: {tail}"));
+            }
+            result.push_str(prefix);
+            text = tail;
+        } else {
+            result.push_str(text);
+            return Ok(result);
+        }
     }
-    Ok(res)
 }
 
 /// Turns a stream of lines into a list of definitions.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -203,7 +203,7 @@ mod test {
                 SYM,
                 |c| !c.contains(EMOJI_PRESENTATION_SELECTOR),
             ) ,
-            "unexpected use of emoji presentation selector (U+FE0F) in `sym` (see list above)",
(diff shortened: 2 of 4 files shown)


T0mstone:
I'll be the first to bring up automation here:
Whether we want to automate this in the future or not, I think we should definitely have all of these hard-coded in the source code, since such automation would probably take a lot of time that we shouldn't have to pay on each build.
The automation would then only be an optional verification step like in #9. (And maybe we always enable it in CI)

That means any concerns about such automation should not block this PR.

MDLC01:
This PR now contains meta changes.

T0mstone:
> For now, I query the list of presentation sequences defined by Unicode from the internet every time the tests are run. Feel free to suggest a better way.

How about caching it in a file `presentation-sequences.txt` and adding `/presentation-sequences.txt` to the `.gitignore`?

Also, I'd like this test to be opt-in to begin with. IMO running `cargo test` shouldn't perform any web requests without the user's consent. A crate feature for this would also have the advantage of making the `reqwest` dependency optional.

T0mstone:
> Additionally, I want to clarify that the dependency on `reqwest` is for tests only.

Yeah I got that, but it's still a huge dep tree that not everyone may want to have to download before running the tests.

MDLC01:
I switched to `ureq`. Additionally, the file is now pinned to Unicode 16.0.0 so as to prevent sudden breakage when a new Unicode version releases. The file is now downloaded in `build.rs`. To prevent always having `ureq` as a build dependency, I hid the tests that require it behind a non-default `_test-unicode-conformance` feature and added it to CI.

MDLC01:
According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies), it is not possible to have a build dependency for tests only, so this feature trick is necessary:

> The same applies to `cfg(debug_assertions)`, `cfg(test)` and `cfg(proc_macro)`. [...] There is currently no way to add dependencies based on these configuration values.

build.rs
```diff
@@ -68,6 +68,18 @@ fn main() {
     let out = std::env::var_os("OUT_DIR").unwrap();
     let dest = Path::new(&out).join("out.rs");
     std::fs::write(&dest, buf).unwrap();
+
+    #[cfg(feature = "_test-unicode-conformance")]
+    std::fs::write(
+        Path::new(&out).join("emoji-variation-sequences.txt"),
+        ureq::get("https://www.unicode.org/Public/16.0.0/ucd/emoji/emoji-variation-sequences.txt")
```
T0mstone, line 75: This still re-dowloads it every time you want to run this test. It might not be much of an issue, but I'd still like it if the file were actually cached in the source tree, especially now that it's pinned to a specific Unicode version, so it doesn't even change usually.
Re-dowloading should only be necessary if the file hasn't been downloaded before or if the cached file is for an older Unicode version (whenever we update that in the future).

T0mstone: I'm still not perfectly happy with this (changes to `src/` shouldn't invalidate the cached file), but I also feel bad for nagging you like this, so I'd be fine with merging as-is and I can just write it myself in a follow-up PR.

MDLC01: To be clear, I pushed a commit that should make the build script only download the file if it is not already in the Cargo cache. So this shouldn't be an issue anymore.

T0mstone: Alright.
The thing I was worried about was that cargo generates a new `OUT_DIR` sometimes (I have 36 of them for codex alone), but I thought it did this way more often than it apparently does: I just tried changing a bunch of stuff in the source code to trigger a new `OUT_DIR` being generated, but I couldn't get cargo to do it, so I'll have to drop that concern now.

src/lib.rs
```diff
@@ -187,4 +190,152 @@ mod test {
...
+    #[cfg(feature = "_test-unicode-conformance")]
+    #[test]
+    fn symbols_have_text_presentation() {
+        let require_presentation_selector = get_valid_presentation_sequences()
+            .into_iter()
+            .map(|s| s.chars().next().unwrap())
+            .collect::<HashSet<_>>();
+        assert!(
+            are_all_variants_valid(SYM, |c| {
+                c.chars().count() != 1
```
T0mstone, line 272: This feels a bit weak; Are you sure there can't be any false positives by just checking that there's more than one codepoint?

MDLC01: I added a comment to clarify that.

src/lib.rs
```diff
@@ -285,6 +287,8 @@ mod test {
             .collect::<HashSet<_>>();
         assert!(
             are_all_variants_valid(EMOJI, |c| {
+                // All text presentations are exactly 2 codepoints long as of
```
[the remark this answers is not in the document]
MDLC01: I changed to "emoji variation sequences" everywhere, as this is how text/emoji presentation sequences seem to be referred to by Unicode.

MDLC01:
Could be `\vs1`–`\vs16`, or alternatively `\evs` and `\tvs` for "emoji variation selector" and "text variation selector" only.[^1]

[^1]: https://www.unicode.org/charts/PDF/UFE00.pdf

MDLC01:
An alternative is to simply paste the corresponding codepoints in the symbol lists

Enivex:
> Or `\emoji` and `\text`

I think this would be good

MDLC01:
Since this does not update the public interface, I won't make a PR to typst/typst just to upgrade the commit hash as the upgrade will happen when the new release is made anyway.

laurmaedje:
> Since this does not update the public interface, I won't make a PR to typst/typst just to upgrade the commit hash as the upgrade will happen when the new release is made anyway.

I think it couldn't hurt to upgrade it already and add a couple of tests for the variant selector stuff. It should fix some stuff, right?

MDLC01:
I'll do it later then

T0mstone:
Just ftr, upgrading the dependency is still blocked on [typst#6489](https://github.com/typst/typst/pull/6489).

laurmaedje:
I wasn't sure whether other people were still gonna leave approvals or not, but it didn't seem like it, so I went ahead and just merged. Thanks!

PR #116, opened 2025-08-07
Move numbering kinds to Codex

MDLC01:
Following https://github.com/typst/typst/pull/6379#issuecomment-2943420158, this PR adds numbering kinds to Codex (renamed to "numeral system" because this is the more appropriate name I believe). This is an initial implementation that may have multiple issues. The goal is to initiate the discussion on this topic.

The current implementation depends on `ecow`, which means the version used here may have to be somewhat synchronized with the one used in Typst. I'm not sure exactly how Cargo determines what version to use, so maybe this is a non-issue.

For now, I did not include the single-character representation of numeral systems as part of this. Should it be in Codex or only in Typst?

For the long names, I chose to use a symbol-like syntax for now (e.g., `chinese.simplified`, `Chinese.simplified`, `chinese.traditional`, and `Chinese.traditional`). This is of course up to debate, and one of the main open question for now. In particular, we might want to align ourselves with [CSS Counter style names](https://www.w3.org/TR/predefined-counter-styles/) as mentioned on Discord [^1]

For now, I only included the formats that were already present in Typst.[^2]

Finally, I realized this could be the occasion to broaden the crate description in `Cargo.toml`. On Discord, @laurmaedje mentioned that "we could expand codex from just naming symbols to more generally supporting Unicode- and internationalization-related efforts in the Typst ecosystem".[^1]

[^1]: https://discord.com/channels/1054443721975922748/1277628305142452306/1380114491638943796
[^2]: https://github.com/typst/typst/blob/bcc71ddb9b5146b97cb3078ecdb5406f31781ea2/crates/typst-library/src/model/numbering.rs


MDLC01:
Other question: some numeral systems simply do not have a way to represent some numbers (e.g., 0 in Roman numerals, or numbers greater than 50 in circled Arabic numerals). Should we change `NumeralSystem::apply`'s return type to `Option<EcoString>` and let the user of Codex handle those cases?

MDLC01:
This is clearly not something that is within the scope of this PR, but numeral systems could be merged with the concept of number formats, making `apply` accept an `f64` and return some string when the numeral system is able to express the number.

MDLC01:
I removed the dependency on `ecow` by using `Display` as you suggested. I did so without a `NumeralSystemSpec` enum because this felt unnecessary, especially because we do not expose the formatting functions.

I'm working on preventing allocations in `positional` and `bijective`.

MDLC01:
I think allocating a vector with `Vec::with_capacity` is better. If we want to prevent allocations, let's do it using an algorithm that works all the time.

MDLC01:
I did not add tests for every numeral system because I would like to go over them in separate PRs to make sure that they are all correctly implemented.

src/numeral_systems.rs
```diff
@@ -0,0 +1,715 @@
...
+    while n != 0 {
+        n -= 1;
+        digits.push(symbols[(n % radix) as usize]);
+        n /= radix;
+    }
+    for digit in digits.iter().rev() {
+        write!(f, "{}", digit)?
+    }
+    Ok(())
+}
```
T0mstone, line 558: The formula for `size` is from Wikipedia, the rest was actually stupidly easy: Just subtract 1111...111 in the given base and apply the positional algorithm.
We probably don't need the special case for `radix = 1`, but I still added it just in case.
```suggestion
fn bijective(f: &mut Formatter<'_>, symbols: &[char], mut n: u64) -> std::fmt::Result {
    if n == 0 {
        return write!(f, "-");
    } else if let [one] = symbols {
        for _ in 0..n {
            write!(f, "{one}")?;
        }
        return Ok(());
    }

    let radix = symbols.len() as u64;
    let size = ((n + 1) * (radix - 1)).ilog(radix);
    n -= (radix.pow(size) - 1)/(radix - 1);
    // For a number of size 1, the MSD's place is the ones place, hence `- 1`.
    let mut most_significant_digit_place = radix.pow(size - 1);
    for _ in 0..size {
        let most_significant_digit = n / most_significant_digit_place;
        write!(f, "{}", symbols[most_significant_digit as usize])?;
        n -= most_significant_digit * most_significant_digit_place;
        most_significant_digit_place /= radix;
    }
    Ok(())
}
```

MDLC01: > Just subtract 1111...111 in the given base and apply the positional algorithm.

With hindsight this is so obvious. Thanks!

The case for base-1 is already not treated properly in `positional` and we don't need it for now, so let's not worry about it for now.

MDLC01:
See also: https://github.com/typst/typst/pull/7632

PR #122, opened 2025-10-25
Add `pee`

MDLC01:
The idea of adding U+2118 ℘ SCRIPT CAPITAL P (with Unicode normative alias WEIERSTRASS ELLIPTIC FUNCTION) has been floating around for a while. To be clear, this is not the same as `cal(P)` or `scr(P)`. This is a different symbol, akin to `ell`. Quoting [UTR #25](https://www.unicode.org/reports/tr25/tr25-15.pdf) (Section 2.3):

> The two characters U+2113  ℓ  SCRIPT SMALL L, and U+2118  ℘  SCRIPT CAPITAL P, are not regular script characters, despite their character names.

The name `pee` is a bit unfortunate, but it is consistent with `ell`. For this reason, I will wait for an approval from @laurmaedje before merging this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### New in `sym`
 
+- Mathematical symbols
+  - `pee`: ℘
+
 - Currency
   - `riyal`: ⃁
 


MDLC01:
I updated the changelog as well

mkorje (approved):

PR #123, opened 2025-10-25
Update to Unicode 17

MDLC01:
This simply updates all mentions of a Unicode version from 16 to 17.

Looking at the [Unicode 17.0.0 changelog](https://www.unicode.org/versions/Unicode17.0.0/), this version does not make any change to presentation sequences, and I also don't think it changes anything related to the `styling` module, so this change is trivial.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -72,7 +72,7 @@ fn main() {
     {
         let emoji_vs_list = Path::new(&out).join("emoji-variation-sequences.txt");
         if !std::fs::read_to_string(&emoji_vs_list)
-            .is_ok_and(|text| text.contains("Emoji Version 17.0"))
+            .is_ok_and(|text| text.contains("Version: 17.0"))
         {
             let content = ureq::get(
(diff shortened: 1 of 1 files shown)


MDLC01:
I updated the changelog to mention the trivial bump to Unicode 17.0.0.

build.rs
```diff
@@ -72,10 +72,10 @@ fn main() {
     {
         let emoji_vs_list = Path::new(&out).join("emoji-variation-sequences.txt");
         if !std::fs::read_to_string(&emoji_vs_list)
-            .is_ok_and(|text| text.contains("Emoji Version 16.0"))
+            .is_ok_and(|text| text.contains("Emoji Version 17.0"))
```
MDLC01:
```suggestion
            .is_ok_and(|text| text.contains("Version: 17.0"))
```

This line is not present is the new version of the file so it needs to be changed.

mkorje (approved):

PR #124, opened 2025-10-25
Add Saudi riyal currency symbol

MDLC01:
This PR adds the symbol for the Saudi riyal currency, which was added to Unicode in version 17.0.0.

I chose the name `riyal` instead of something more specific because this is the only riyal in active use.[^1]

[^1]: According to @Enivex on Discord: https://discord.com/channels/1054443721975922748/1277628305142452306/1431770363737407562

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -741,6 +741,7 @@ peso $
   .philippine ₱
 pound £
 riel ៛
+riyal ⃁
 ruble ₽
 rupee
   .indian ₹


Enivex:
I should say. There are multiple currencies called riyal/rial, but to my knowledge the alternative U+FDFC ﷼ is not in active use for any of them.

Edit: 

https://www.unicode.org/L2/L2025/25076-sew-recommendations-saudi-riyal.pdf

> U+FDFC RIAL SIGN was encoded in Unicode 3.2 as a compatibility character for Iranian standards that defined such
a character

mkorje (approved):

PR #126, opened 2025-10-26
Add tests for validity of standardized variation sequences

MDLC01:
This PR updates the testing infrastructure to also test that standardized variation sequences are valid. I also reorganized `build.rs` to separate the processing of Codex module files from the part that downloads Unicode data files for the tests.

Standardized variation sequences are defined by Unicode in [`StandardizedVariants.txt`](https://www.unicode.org/Public/UCD/latest/ucd/StandardizedVariants.txt) and are sequences consisting of an initial character and a variation selector from VS1 to VS14. Notably, standardized variation sequences are disjoint from emoji variation sequences (i.e., presentation sequences), which we already fully support since https://github.com/typst/codex/pull/114.

Unicode defines a third kind of variation sequences: ideographic variation sequences, which we do not care about in Codex.

More on variation sequences: https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-23/#G26678.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,39 +191,10 @@ mod test {
         }
     }
 
-    /// https://www.unicode.org/reports/tr51/#def_text_presentation_selector.
-    const TEXT_PRESENTATION_SELECTOR: char = '\u{FE0E}';
-    /// https://www.unicode.org/reports/tr51/#def_emoji_presentation_selector.
-    const EMOJI_PRESENTATION_SELECTOR: char = '\u{FE0F}';
-
-    #[test]
-    fn symbols_are_not_emojis() {
-        assert!(
-            are_all_variants_valid(
-                SYM,
-                |c| !c.contains(EMOJI_PRESENTATION_SELECTOR),
-            ) ,
-            "unexpected use of emoji presentation selector in `sym` (see list above)",
-        )
-    }
-
-    #[test]
-    fn emojis_are_not_text() {
-        assert!(
-            are_all_variants_valid(
-                EMOJI,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -242,13 +313,7 @@ mod test {
...
         assert!(
             are_all_variants_valid(ROOT, |c| {
-                if c.contains(TEXT_PRESENTATION_SELECTOR)
-                    || c.contains(EMOJI_PRESENTATION_SELECTOR)
-                {
-                    sequences.contains(c)
-                } else {
-                    true
-                }
+                !is_presentation_sequence(c) || sequences.contains(c)
```
Akari202: By my understanding c can be a a zero width joiner sequence and contain multiple valid presentation sequences within it causing this test to fail. I know this was already an issue and might be out of scope for this pr (in which case I'm happy to submit one seperately) but I think checking a window over c is slightly more accurate. 

"woman walking facing right: light skin tone" (1F6B6 1F3FB 200D 2640 FE0F 200D 27A1 FE0F ) for instance fails despite 2640 FE0F and 27A1 FE0F being individually valid and contained in sequences

MDLC01: You are right. The right fix, since there is no definitive list of all valid emoji sequences (see https://github.com/typst/codex/pull/167#issuecomment-4450156364) would probably be to get all presentation sequences from the symbol and check that they are individually valid.

MDLC01: I pushed a fix which actually detected two missing presentation selectors.

MDLC01:
Tests are failing because CI is using a version of Rust older than 1.91.0

PR #132, opened 2025-12-18
Add narrow emptyset symbol

mkorje:
I've gone with the modifier `.alt` for now, matching things like `pi.alt` etc. Please share any other ideas you might have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### New in `sym`
 
 - Mathematical symbols
+  - `emptyset.zero`: ∅︀
+  - `nothing.zero`: ∅︀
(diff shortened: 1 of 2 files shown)


MDLC01:
> It's widely regarded as a mistake

Do you have sources about that? If this is true and the narrow variant originates from a mistake and isn't used much, then I would agree with not making it more accessible.

> It's not like #127 makes the symbols `inter.alt` and `union.alt`.

My feeling is that the narrow emptyset variant is really _the_ alternative form of the emptyset variant, while the serif variants of union and intersection symbols are somewhat more niche and specific. But I don't have a strong opinion and I'm fine with `.zero` if other people agree that this is better than `.alt`.

PR #133, opened 2025-12-19
Add more `.not` variants

MDLC01:
Some mathematical characters do not have a precomposed "not" variant. Instead, Unicode allows using U+0338 COMBINING LONG SOLIDUS OVERLAY to produce the negation of a character.

In this pull request, I added a few of those decomposed forms. More specifically, I added the ones that are listed in Table 2.8 of [UTR \#25](https://www.unicode.org/reports/tr25/) ([Table 9 in the recent proposed update](https://www.unicode.org/reports/tr25/tr25-16.html#table_using_vertical_line)) for which the base characters already have a name in Typst. I decided to stick to those for now as a starting point, but I definitely think we should add `.not` variants for all relations in the future.

Note that U+20D2 COMBINING LONG VERTICAL LINE OVERLAY is also a valid way to encode negation, but serves as an alternative to U+0338. If we decide to add them in the future, they would not be using `.not` (possible options include `.not.alt` or `.neg`).

Regarding font support, none of the few fonts I tested[^1] seem to support this for now in the sense that there does not seem to be specific glyphs for those symbols. Below is the result I get with New Computer Modern Math. We may want to reach out to the maintainer to make them aware of this Unicode feature.

The combining slash is not placed very gracefully but the intent is still understandable.

[^1]: New Computer Modern Math, Fira Math, Libertinus Math and Lete Sans Math.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 ### New in `sym`
 
 - Mathematical symbols
+  - `in.small.not`: ∊̸
+  - `in.rev.small.not`: ∍̸
+  - `minus.tilde.not`: ≂̸
+  - `eq.quad.not`: ≣̸
+  - `lt.equiv.not`: ≦̸
+  - `gt.equiv.not`: ≧̸
+  - `lt.eq.slant.not`: ⩽̸
+  - `gt.eq.slant.not`: ⩾̸
   - `gt.double.nested`: ⪢
   - `lt.double.nested`: ⪡
   - `gt.arc` ⪧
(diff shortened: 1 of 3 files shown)


Enivex:
In my opinion it would make more sense to introduce a function for this, like we do for combining accents. This can be used for all symbols.

There's a bunch more we can add to (though font support is lacking). At one point I started writing down names, but didn't get very far

https://en.wikipedia.org/wiki/Combining_Diacritical_Marks

https://en.wikipedia.org/wiki/Combining_Diacritical_Marks_for_Symbols

MDLC01:
I want to get back to this, but for now I will close this PR to reduce the amount of open PRs and because New Computer Modern Math does not support those combining sequences anyway.

PR #146, opened 2026-02-25
Add dots.h.low as alias for .h

knuesel:
It seems that having `dots` default to the centered version was meant to be implemented already in https://github.com/typst/typst/pull/747 but that implementation was wrong (just changing the order of .h and .h.c without considering the number of modifiers).

Closes https://github.com/typst/codex/issues/147

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -186,8 +186,9 @@ slash /
   .triple ⫻
   .big ⧸
 dots
-  .h.c ⋯
   .h …
+  .h.c ⋯
+  .h.low …
   .v ⋮
   .down ⋱
   .up ⋰


MDLC01:
Since I expect `dots` to be commonly used in math, I would suggest the following path that ensures a smooth transition:
- In the next version (i.e., this PR), introduce `dots.h.b` and deprecate `dots.h`, suggesting to use `dots.h.b` instead.
- In a future version, change `dots.h` to be the centered version and deprecate `dots.h.c`, suggesting to use `dots.h` instead.
- In a later version, completely remove `dots.h.c`.

That way, we ultimately end up with `dots.h` and `dots.h.b` and no `dots.h.c`.

Unrelatedly, I don't love the `.b` modifier here, but I don't have a better solution so it's probably fine. An alternative is to rename the current `dots.h` to `ellipsis`, but I think this is just confusing.

knuesel:
Good idea. It wasn't clear to me this deprecation was possible but l guess there's no problem if we have two different lines for the same symbol.

I have the same feeling for `.b`. Another option would be to use `dots.op` for `dots.h.c` and `dots.basic` for `dots.h.b`. This would mirror the relationship between `dot.op` and `dot.basic`.

PR #148, opened 2026-03-05
Add chess symbols

MDLC01:
This supersedes https://github.com/typst/codex/pull/73.

I simply added chess symbols from Unicode 17.0 and updated the changelog. I also fixed a typo. Apart from that, the content is that same as https://github.com/typst/codex/pull/73.


MDLC01:
I extracted part of this to https://github.com/typst/codex/pull/176. I think it will be easier to get this merged if we do it in smaller pieces.

Issue #95, opened 2025-07-03
Allow the use of order-dependence to disambiguate symbol variants


Enivex:
Let me first say: This is _not_ a proposal to make symbol names order-dependent in general. _Only_ variants where order is needed for disambiguation would be order-dependent.

Currently, ⪋ and ⪌ are `lt.gt.equiv` and `gt.lt.equiv`. However, the symbols `⪑` and `⪒` also exist. In order to name this in our current scheme, one would need to invent yet more unnatural modifiers.

With this proposal, we would be able to distinguish between

`lt.equiv.gt` ⪋
`lt.gt.equiv` ⪑
`gt.equiv.lt` ⪌
`gt.lt.equiv` ⪒

It would also enable https://github.com/typst/codex/pull/75 without abusing submodules, assuming we also get https://github.com/typst/codex/issues/94.

(Not to mention there are examples of current naming that could potentially be improved with this proposal, such as for arrows.)

knuesel:
Thinking about what the alternatives would be, we could put these symbols under `eq` and add a modifier for the position of the equal sign:

```
```
eq.lt.gt ⪋
eq.lt.gt.bottom ⪑
eq.gt.lt ⪌
eq.gt.lt.bottom ⪒
```

though it's maybe not obvious what "bottom" applies to.

Enivex:
> Thinking about what the alternatives would be, we could put these symbols under `eq` and add a modifier for the position of the equal sign:
> 
> ```
> eq.lt.gt ⪋
> eq.lt.gt.bottom ⪑
> eq.gt.lt ⪌
> eq.gt.lt.bottom ⪒
> ```
> 
> though it's maybe not obvious what "bottom" applies to.

If all we care about is giving unique names, then yes, but this completely loses the meaning of each symbol.

knuesel:
and it's just wrong, I mean `eq.lt.gt` and `eq.gt.lt` cannot both exist if order doesn't matter (I don't know what I was thinking).

Enivex:
That too

Enivex:
We could also use this to potentially get rid of `neq`, `nequiv`, `napprox` and `ntilde`.

As in `lt.eq.not` vs `lt.not.eq`

MDLC01:
I don't think this would be a good idea. `.neq` is fine imo.

[...]

Issue #94, opened 2025-07-03
Allow repeated modifiers to simplify naming


Enivex:
For a handful of symbols, it would be useful to be able to repeat modifiers. I give two examples here.

1) We currently have

× - `times`
⊗ - `times.o`

Another variant that has not been named yet is ⨷, where `times.o.o` is an obvious name candidate.

2) In "Supplemental Mathematical Operators" one finds ⪓ and ⪔.

Natural names would be `lt.seq.gt.seq` and `gt.seq.lt.seq`.

(Or any of a number of other alternatives, but they all share the need to repeat one or more modifiers)

This is only somewhat orthogonal to making it possible to disambiguate using order (the other proposal)

PR #151, opened 2026-03-26
Add missing tacks

MDLC01:
- `tack.r.triple` was simply missing for no reason.
- `ttack.r` and `ttack.r.not` were already present as `forces` and `forces.not`. This adds all the other similar variants under non-semantic names because I don't think there is a good semantic name.
- There is no non-short version of `ttack.t.short` and `ttack.b.short` in Unicode AFAIK. `ttack.t` and `ttack.b` fall back to `ttack.t.short` and `ttack.b.short` respectively.
- The `.double` and `.triple` modifiers on `tack` are already used for something else, hence the proposed `ttack` and `tttack` symbols. Those names aren't the greatest but they are the best I was able to come up with. Feel free to make other suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,8 +62,8 @@
   - `bowtie.filled.l`: ⧑
   - `bowtie.filled.r`: ⧒
   - `tack.rrr`: ⫢
-  - `tacks.ll.rr`: ⟚
-  - `tacks.t.b.short`: ⫩
+  - `tacks.llrr`: ⟚
+  - `tacks.tb`: ⫩
 
 - Currency
   - `riyal`: ⃁
@@ -85,7 +85,7 @@
 - `lt.tri` and variants in favor of `lt.closed`
 - `join` and its variants in favor of `bowtie.big` with the same variants
 - `tack.r.double`, `tack.r.double.not`, `tack.l.double`, `tack.t.double`, and `tack.b.double` in favor of `tack.rr`, `tack.rr.not`, `tack.ll`, `tack.tt`, and `tack.bb` respectively.
-- `tack.l.r` in favor of `tacks.l.r`.
+- `tack.l.r` in favor of `tacks.lr`.
 
 ### Removals **(Breaking change)**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,16 @@
   - `bowtie.filled`: ⧓
   - `bowtie.filled.l`: ⧑
   - `bowtie.filled.r`: ⧒
+  - `tack.r.triple`: ⫢
+  - `ttack.r`: ⊩
+  - `ttack.r.not`: ⊮
+  - `ttack.r.double`: ⊫
+  - `ttack.r.double.not`: ⊯
+  - `ttack.l`: ⫣
+  - `ttack.l.double`: ⫥
+  - `ttack.t.short`: ⫨
+  - `ttack.b.short`: ⫧
+  - `tttack.r`: ⊪
 
 - Currency
   - `riyal`: ⃁
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,8 +61,8 @@
   - `bowtie.filled`: ⧓
   - `bowtie.filled.l`: ⧑
   - `bowtie.filled.r`: ⧒
-  - `tack.ll.rr`: ⟚
-  - `tack.t.b.short`: ⫩
+  - `tacks.ll.rr`: ⟚
+  - `tacks.t.b.short`: ⫩
 
 - Currency
   - `riyal`: ⃁
@@ -84,6 +84,7 @@
 - `lt.tri` and variants in favor of `lt.closed`
 - `join` and its variants in favor of `bowtie.big` with the same variants
 - `tack.r.double`, `tack.r.double.not`, `tack.r.triple`, `tack.l.double`, `tack.t.double`, and `tack.b.double` in favor of `tack.rr`, `tack.rr.not`, `tack.rrr`, `tack.ll`, `tack.tt`, and `tack.bb` respectively.
+- `tack.l.r` in favor of `tacks.l.r`.
 
 ### Removals **(Breaking change)**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,16 +61,8 @@
   - `bowtie.filled`: ⧓
   - `bowtie.filled.l`: ⧑
   - `bowtie.filled.r`: ⧒
-  - `tack.r.triple`: ⫢
-  - `ttack.r`: ⊩
-  - `ttack.r.not`: ⊮
-  - `ttack.r.double`: ⊫
-  - `ttack.r.double.not`: ⊯
-  - `ttack.l`: ⫣
-  - `ttack.l.double`: ⫥
-  - `ttack.t.short`: ⫨
-  - `ttack.b.short`: ⫧
-  - `tttack.r`: ⊪
+  - `tack.ll.rr`: ⟚
+  - `tack.t.b.short`: ⫩
 
 - Currency
   - `riyal`: ⃁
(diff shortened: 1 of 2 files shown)


MDLC01:
There is also U+2AE6 ⫦, which I don't know how to name.

Enivex:
What's the motivation behind ttack and tttack? It seems a bit awkward to me

isuffix:
I also dislike `ttack` and `tttack`, they feel like the kind of LaTeX symbol names which are clever, but are hard to remember and discover, like `\oint`. It's also uncomfortable to me that the existing `tack.r.l` (⟛) doesn't have the same base as `ttack.l` (⫣) or `ttack.r` (⊩).

**Suggestion:**
What if we make more use of repeated letters instead of `.double/.triple`? If we use `.v` as the letter for the vertical stem of the tack, then we could assign them like this:
```txt
tack
  .v.r ⊢
  .v.rr ⊨
  .vv.r ⊩
  .vv.rr ⊫
```
If we order the `.v` variants before `.vv` (and keep `.double` as an alias to `.rr`, `.ll`, etc. for a few versions), fallback to the existing tacks should work alright. And for the up/down facing tacks, we could either use `.h` to refer to the horizontal base or keep using `.v` for consistency (or maybe use `.s` for both as the 'stem' of the tack?), I'm not sure which would be best.

<details><summary>Here's that expanded to all the tacks (using just <code>.v/.vv/.vvv</code>)</summary>

```txt
tack
  .v.r ⊢
  .v.r.not ⊬
  .v.r.long ⟝
  .v.r.short ⊦
  .v.rr ⊨
  .v.rr.not ⊭
  .v.rrr ⫢
  .vv.r ⊩
  .vv.r.not ⊮
  .vv.rr ⊫
  .vv.rr.not ⊯
  .vvv.r ⊪
  .v.l ⊣
  .v.l.long ⟞
  .v.l.short ⫞
  .v.ll ⫤
  .vv.l ⫣
  .vv.l.r ⟛
  .vv.ll ⫥
  .v.t ⊥
  .v.t.big ⟘
  .v.t.short ⫠
  .v.tt ⫫
  .vv.t.short ⫨
  .v.b ⊤
  .v.b.short ⫟
  .v.b.big ⟙
  .v.bb ⫪
  .vv.b.short ⫧
```

</details>

isuffix:
For U+2AE6 ⫦ I feel like `.split` or `.apart` may be good modifiers. E.g. `tack.r.double.split` currently or `tack.v.rr.split` under my proposal above.

MDLC01:
> What's the motivation behind ttack and tttack? It seems a bit awkward to me

I agree but I did not have any better idea at the time.

> What if we make more use of repeated letters instead of `.double/.triple`?

I agree this is a better idea, but I don't like the use of the `.v` modifier. What about the following?

```
tack
  .r ⊢
  .rr ⊨
  .doube.r ⊩
  .double.rr ⊫
```

This would be a breaking change, however.

> For U+2AE6 ⫦ I feel like `.split` or `.apart` may be good modifiers.

Could you provide a rationale for those names? I'm not sure exactly what is "split" or "apart" in this symbol.

> It's also uncomfortable to me that the existing `tack.r.l` (⟛) doesn't have the same base as `ttack.l` (⫣) or `ttack.r` (⊩).

I should mention that there are two other similar symbols in Unicode that don't yet have a name in Codex: U+2AE9 ⫩ and U+27DA ⟚. I should probably add them in this PR as well as they are pretty straight forward to name given the precedent of `tack.r.l`. Unicode does not define any other tack symbol as far as I am aware.[^1]

[^1]: There are two other symbols containing up tacks, but they are more specifically perpendicular symbols: U+29B9 ⦹ and U+2AE1 ⫡.

isuffix:
Keeping `.rr`, but using `.double`/`.triple` for the stem is a nice compromise! I'm in favor.

For `.split`/`.apart` I think I misconstrued what the symbol looks like and checking it again I also don't really know what I was thinking. My rationale was something to the effect that in U+2AE6 ⫦ there's a larger/more prominent visual gap between the two lines than in, say U+22A9 ⊩, but I don't really see that now.

Maybe a name for U+2AE6 ⫦ could be `tack.r.mid`, since it has an extra line in the middle of the tack?

MDLC01:
I added names for U+2AE9 ⫩ and U+27DA ⟚, and switched to use repeated letters instead. This makes this PR a breaking change.

Regarding ⟛, ⫩, and ⟚, we may want to use `tacks` as the base symbol for consistency with `arrows` and `harpoons`, but I don't think this is necessary because the family is quite small (only three variants).

knuesel:
I would prefer `tacks` for ⟛, ⫩, and ⟚ as `.l.r` and `.t.b` evoke "modifications" in both directions rather than duplication.

I guess the double/triple variants are omitted so we can show the deprecations for one release. Should there be a sibling draft PR to keep track of the symbols that don't make it here, so we don't forget?

For U+2AE6 ⫦ what about `tack.double.r.through` ?

MDLC01:
> I guess the double/triple variants are omitted so we can show the deprecations for one release. Should there be a sibling draft PR to keep track of the symbols that don't make it here, so we don't forget?

Exactly. We will just have to go back to this PR when removing the deprecations in the next version. I don't want to open the new PR right now to prevent having too many open PRs.

Enivex:
This issue illustrates how naming would be much simpler with #94 and #95 I think

MDLC01:
In this specific case I don't think we need either.

isuffix:
> In this specific case I don't think we need either.

I assume Enivex is implying that we could replace `.rr` with `.r.r`, although to me that feels a bit too verbose.

Enivex:
> > In this specific case I don't think we need either.
> 
> I assume Enivex is implying that we could replace `.rr` with `.r.r`, although to me that feels a bit too verbose.

I was primarily thinking of disambiguating the different uses of double

MDLC01:
To summarize the current state of this PR:
- There seems to be a consensus on the change from `.double` and `.triple` to repeated letters (e.g., `.rr` and `.rrr`), and in a later version introduce ⊩ & co. using the newly available `.double` and `triple` modifiers.
- There is no consensus for U+2AE6 ⫦.
    - Let's wait for the PR that introduces ⊩ before worrying about that, so that we can move forward with this PR.
- The remaining open question is whether to use `tack` or `tacks` as the base symbol for ⟛, ⫩, and ⟚.
    - There is precedent for the plural form with `arrows`.
    - I don't think anyone expressed an opinion against the plural form.
    - I am personally in favor of `tacks` so I added a commit to change that.

knuesel:
I like `tacks`, but it's unfortunate that `tacks.l` would default to `tacks.l.r` ⟛.

(With `arrows` it's a bit different: we have `.lr` and `.rl` which also distinguishes the ordering.)

What about `tacks.h` for ⟛, `tacks.hh` for ⟚ and `tacks.v` for ⫩ ?

isuffix:
I'm still preferential to the `tack.l.r` forms instead of `tacks`. I think the issue of `tacks.l` highlights that they aren't paired in the same way as arrows or harpoons. With those, the symbols are always stacked above or next to each other, but the paired tacks are always facing away from eachother back-to-back, and that feels different to me.

I think `tacks.h` could be a good resolution to that, but I don't like it in the context of all of the other tacks using `.{l,r,t,b}`. The difference feels unecessarily confusing, and going with `tack.l.r` fits in better with the existing pattern.

I also hate that `tacks.short` would not be horizontal like `tack.short` since we don't have any paired horizontal short tacks.

T0mstone:
I think the consistency with `arrows` and `harpoons` is worth it to have `tacks.lr`. In theory, there could be a symbol like `tacks.rl` with the two tacks facing towards each other, it's just incidental that Unicode doesn't have it.

knuesel:
Going with `tacks.lr` for ⟛, we would use `tacks.llrr` for `⟚`  and `tacks.tb` for ⫩ ? That would be OK for me.

I don't think need `tacks.short` at all, what's the rationale for it?

MDLC01:
> I don't think need `tacks.short` at all, what's the rationale for it?

⫩ is intended to be a short up tack above a short down tack, hence the `.short` modifier.

But since this is the only instance of an up tack above a down tack, we can probably remove the `.short` modifier. This would solve one of @isuffix's concerns.

MDLC01:
I changed `tacks` to use the names suggested by @knuesel.

If we can't have a consensus on that, I'll just remove this change from the PR so that what we agree on can be merged.

MDLC01:
I reverted the deprecation of `tack.l.r` and removed `tacks`. We should have a consensus on what remains.

isuffix (approved):

T0mstone (approved):

sijow (approved):

MDLC01:
There have been two weeks for anyone to raise any concern, so we can merge this immediately.

PR #153, opened 2026-04-07
Add control pictures

zacharie-moughanim:
The "whitespace glyph" is the symbol often used to designate the spacebar on keyboards : ␣.
I'm honestly not sure about the name, nor if it is pertinent to put it as a variant of `space`, which always designate whitespaces.


MDLC01:
Note that a similar character exists in Unicode: U+2420 ␠ SYMBOL FOR SPACE. Both come from the [Control Pictures](https://unicode.org/charts/PDF/U2400.pdf) block.

Enivex:
The name in html is `blank`, but https://www.compart.com/en/unicode/U+2422 also exists

zacharie-moughanim:
Maybe we could remove forget about the name `space.glyph`, add the name `blank` and group ␣, U+2420, U+2422 as variants of this name, what do you think ?

MDLC01:
According to French Wikipédia, U+2422 ␢ is used in theoretical computer science.[^1] I am struggling to find uses of this symbol other than the book mentioned on this page, and unable to locate the proposal that lead to its inclusion to Unicode. I do not know whether we actually want to assign it a name.

My intuition is that the graphic pictures for control codes (i.e., 2400..2421 and 2424) should be in their own sub-module because they are so clearly related without being variants of each others. Then, U+2423 ␣ could either be a variant of U+2420 ␠ and U+2422 ␢ `sym.blank`, or U+2423 ␣ could be its own symbol outside of the sub-module but that causes a potential collision if we want to include U+2422 ␢ as well.

[^1]: https://fr.wikipedia.org/wiki/%C9%83#Symbole_blanc

MDLC01:
Also, don't forget to update the changelog to mention the addition(s).

zacharie-moughanim:
Just added each symbol representing a non-printable character under a submodule `control` :
```
␀ // symbol for null
␁ // symbol for start of heading
␂ // symbol for start of text
␃ // symbol for end of text
␄ // symbol for end of transmission
␗ // symbol for end of transmission block
␅ // symbol for enquiry
␇ // symbol for bell
␇ // symbol for backspace
␉ // symbol for horizontal tabulation
␋ // symbol for vertical tabulation
␊ // symbol for line feed
␌ // symbol for form feed
␍ // symbol for carriage return
␎ // symbol for shift out
␏ // symbol for shift in
␐ // symbol for datalink escape
␑ // symbol for device control one
␒ // symbol for device control two
␓ // symbol for device control three
␔ // symbol for device control four
␆ // symbol for acknowledge
␕ // symbol for negative acknowledge
␖ // symbol for synchronous idle
␘ // symbol for cancel
␙ // symbol for end of medium
␚ // symbol for substitute
␦ // symbol for substitute form two
␛ // symbol for escape
␜ // symbol for file separator
␝ // symbol for group separator
␞ // symbol for record separator
␟ // symbol for unit separator
␠ // symbol for space
␡ // symbol for delete
␥ // symbol for delete form two
␢ // blank symbol
␣ // open box
␤ // symbol for newline
```
I got most of the names from [compart](https://www.compart.com/en/unicode/U+241A) and from [this wikipedia page](https://en.wikipedia.org/wiki/Control_character).

Enivex:
I think we should use the full words `acknowledge`, `substitute`, `synchronous`, `escape`, and `delete`

Edit: Missed a couple

zacharie-moughanim:
> I think we should use the full words `acknowledge`, `substitute`, `synchronous`, `escape`, and `delete`
> 
> Edit: Missed a couple

Done

src/modules/sym.txt
```diff
@@ -1268,3 +1268,56 @@ gender {
...
+    .alt ␦
+  escape ␛
+  separator
+    .file ␜
+    .group ␝
+    .record ␞
+    .unit ␟
+  space ␠
+  delete ␡
+    .alt ␥
```
MDLC01: I think we can leave this one out of this PR as this is a symbol used on keyboards rather than a control picture.

src/modules/sym.txt
```diff
@@ -1268,3 +1268,56 @@ gender {
...
+  blank ␢
+    .box ␣
```
MDLC01: Those two definitely belong outside of the `control` submodule imo.

src/modules/sym.txt
```diff
@@ -1268,3 +1268,56 @@ gender {
...
+}
```
MDLC01: Please make sure the file ends with a proper line ending.

MDLC01 (changes requested):
For the control code pictures, I am wondering whether it would actually be better to use the common abbreviations for the symbols (e.g., as listed at https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)#Table_of_characters), which seems to match the letters displayed on the corresponding glyphs?

Right now, multiple variants feel weird, such as `heading.start` or `transmission.end`, because they seem to imply the existence of a dual variant (`heading.end` and `transmission.start`) that does not in fact exist (or rather, it exists under a different name: `heading.end` is `text.start` and `transmission.start` is `heading.start`). I feel like most control codes exist as is and aren't variants of each others, so using separate names would make more sense.

src/modules/sym.txt
T0mstone, line 1: I have to say I don't like this breaking words into modifiers just for the sake of it, especially when there's no meaningful default variant.

zacharie-moughanim:
Replaced each name from the control submodule to abbreviations.
Therefore, I only left the abbreviations in this submodule.
I put the two symbols that are used as keyboard labels out of the module, under the names `spacebar` for `␣` and `delete` for `␥`.
Two symbols from the control pictures block remains : `␦` and `␢`, but they seemingly does not fit within the control submodule, and are not used in keyboard labels, so I left them out of this pr.

zacharie-moughanim:
Also, there are four symbols whose name is not the exact abbreviations, device controls (1, 2, 3, 4) since apparently numbers aren't allowed in symbol names.
I looked at the name for dice, so I named them `dc.one`  ␑, `dc.two`  ␒, `dc.three`  ␓, `dc.four` ␔.
The default is `dc.one` which is arbitrary. Tell me if there is a better way to name them.

src/modules/sym.txt
```diff
@@ -1268,3 +1272,43 @@ gender {
...
+  cr ␍ // Carriage Return
+  so ␎ // Shift Out
+  si ␏ // Shift In
+  dle ␐ // Data Link Escape
+  dc // Device Control
+    .one ␑
+    .two ␒
+    .three ␓
+    .four ␔
+  nak ␕ // Negativ AcKnowledge
```
```suggestion
  nak ␕ // Negative AcKnowledge
```

CHANGELOG.md
```diff
@@ -65,6 +65,47 @@
 - Currency
   - `riyal`: ⃁
 
+- Keyboard labels
+  - `spacebar` : ␣
```
MDLC01:
```suggestion
  - `spacebar`: ␣
```

Same for all the following ones.

zacharie-moughanim: done 3510ebaad43c01ca5f0bc39434935f50b146b986

src/modules/sym.txt
zacharie-moughanim: done 3510ebaad43c01ca5f0bc39434935f50b146b986

MDLC01 (approved):

knuesel:
I wonder if `delete` and `spacebar` at the top level is the best idea, especially for `delete` which sounds a bit generic. There are many other keyboard key symbols to add eventually like U+2BB8 (caps lock), U+2B7E (tab key), U+23CE (return) (with the usual mess of alternatives and variants). Would it make sense to use another submodule for those?

Otherwise this looks good to me.

zacharie-moughanim:
My feeling is that those symbols are referring to `delete` and `space` in general and that the fact they are used as keyboard labels is not prominent enough to put them in a `keyboard` submodule.
The reason for the name `spacebar` and the `// Keyboard` comment is because `space` is already used to denote the control code ` ` and variants, but these symbols are printable version of delete and space.
Plus, I feel like such a submodule could gather a wide range of symbol and I wouldn't know where to draw the line. For instance, it should probably contain `arrow.stroked.t` (SHIFT key), ↵ (currently nameless), and `arrow.l.hook` (RETURN key) but it feels weird to me; I think in most cases these symbols have more meaningful name.
Maybe we could leave these two out of this PR and open a wider issue about what to do with keyboard labels, tell me what you think.

MDLC01:
> especially for `delete` which sounds a bit generic

Thinking about it again, it's true that users might expect `delete` to be something like U+232B ⌫ ERASE TO THE LEFT rather than this obscure symbol. Let's remove it from this PR I think.

> Plus, I feel like such a submodule could gather a wide range of symbol and I wouldn't know where to draw the line. For instance, it should probably contain `arrow.stroked.t` (SHIFT key), ↵ (currently nameless), and `arrow.l.hook` (RETURN key) but it feels weird to me

I don't see a fundamental issue with having ⇧ as `keyboard.shift`.[^1] Regarding the Return key, Unicode has other symbols that more specifically correspond to this (U+23CE ⏎, U+2B90 ⮐, U+2B91 ⮑) so this would be even less of an issue.

---

For now, I think let's remove `delete` and only keep `spacebar` as this was the initial goal of this PR. `spacebar` is probably the best name we can find anyway, and we can deprecate it if/when we add a `keyboard` submodule.

[^1]: In practice, I think it is too opinionated to suggest that this is _the_ symbol for Shift when other arrows in Unicode might correspond better. But this is besides the point of duplicating some symbols in a `keyboard` module.

zacharie-moughanim:
I removed delete from the pr as suggested

>     1. In practice, I think it is too opinionated to suggest that this is _the_ symbol for Shift when other arrows in Unicode might correspond better. But this is besides the point of duplicating some symbols in a `keyboard` module. [↩](#user-content-fnref-1-d1e17e8c015968d9d539385acf7c3323)

I got these examples from wikipedia ([shift key](https://en.wikipedia.org/wiki/Shift_key#Labeling), [return key](https://en.wikipedia.org/wiki/Enter_and_return_keys#Keyboard_symbols)) but I admit I didn't investigate further. I guess this should be thoroughly discussed in an issue regarding keyboard labels.

MDLC01 (approved):

T0mstone (approved):

knuesel (approved):

MDLC01:
Thanks everyone!

laurmaedje:
The changelog entries don't make clear that this lives in a `control` module. That would be good to change.

MDLC01:
> The changelog entries don't make clear that this lives in a `control` module. That would be good to change.

I addressed this in https://github.com/typst/codex/pull/172

PR #154, opened 2026-04-18
Convert `dotless` to a submodule

T0mstone:
This makes more conceptual sense to me, since `dotless` on its own is meaningless.

This is a breaking change, so three approvals are necessary.

Note that there is no way to do a deprecation for this, but I don't think anyone was using `dotless` without modifiers anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,9 @@ These previously deprecated items were removed:
 - `shin`
 - `planck.reduce`
 
+The `dotless` symbol was converted to a module,
+meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
+
 ## Version 0.2.0 (October 7, 2025)
 
 ### General changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   meaning you can no longer use `dotless` as a shorthand for `dotless.i`. **(Breaking change)**
 
 ### Removals in `sym` **(Breaking change)**
+
 These previously deprecated items were removed:
 - `gt.tri.*`
 - `lt.tri.*`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### General changes
+
+- The `dotless` symbol was converted to a module,
+  meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
+
 ## Version 0.3.0 (June 4, 2026)
 
 ### General changes
@@ -192,9 +199,6 @@ These previously deprecated items were removed:
 - `dalet`
 - `shin`
 - `planck.reduce`
-
-The `dotless` symbol was converted to a module,
-meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
 
 ## Version 0.2.0 (October 7, 2025)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### General changes
 
 - The `dotless` symbol was converted to a module,
-  meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
+  meaning you can no longer use `dotless` as a shorthand for `dotless.i`. **(Breaking change)**
 
 ## Version 0.3.0 (June 4, 2026)
 
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -1234,9 +1234,10 @@ pee ℘
 planck ħ
 Re ℜ
 Im ℑ
-dotless
-  .i ı
-  .j ȷ
+dotless {
+  i ı
+  j ȷ
+}
 
 // Miscellany.
 die


Enivex:
I really don't see the issue with having no top level symbol.

Naming will always be a result of compromises.

If someone truly wishes to write `dotless` instead of `dotless.i`, then let them.

MDLC01:
I think the change I would like instead is to change `dotless.i` and `dotless.j` to `i.dotless` and `j.dotless` instead. Especially because some symbols that we may want to add in the future are also variation of Latin letters that would benefit from this naming schema (e.g., `F.inv` for U+2132 Ⅎ, `G.inv` for U+2141 ⅁, `L.inv` for U+2142 ⅂, `L.rev` for U+2143 ⅃).

Of course as of now those would not be accessible in Typst math, but I think it may be sensible to allow the base identifier of a field access to be a single letter in math (the same way the identifier after the period can already be a single letter). @isuffix thoughts?

Enivex:
I think single letter identifiers would be a bad idea.

Something like
```
and therefore
$
 a = b + c.
$
```
would now presumably be a syntax error.

Your examples could (almost) equally well be `inv.F`, `inv.G`, `inv.L`, `rev.L`, with the added benefit that all such characters would be grouped.

T0mstone:
> Something like
> 
> ```
> and therefore
> $
>  a = b + c.
> $
> ```
> 
> would now presumably be a syntax error.

Well, `$ alpha + beta = gamma. $` is not a syntax error today, so I wouldn't think so.

Enivex:
Fair enough, but I still strongly dislike the idea.

T0mstone:
Well, I strongly dislike having things that work like modifiers before the base symbol name. It's the opposite of consistent.

Enivex:
> Well, I strongly dislike having things that work like modifiers before the base symbol name. It's the opposite of consistent.

It's impossible to achieve complete consistency. That's why I mentioned compromises above.

And even if you _could_ achieve perfect consistency, that may not necessarily be the most usable option.

knuesel:
I also think single-letter identifiers are probably a bad idea and that `i` in math should be either a letter or a symbol binding, but not "both, it depends...".

@Enivex I can see that the current state can seem no big deal but is there a downside to the proposed change (other than breaking the code of someone that writes `dotless` instead of `dotless.i`)? It does seem conceptually more correct...

MDLC01:
> I also think single-letter identifiers are probably a bad idea and that `i` in math should be either a letter or a symbol binding, but not "both, it depends...".

A single letter can already be interpreted as an identifier, such as `i` in `dotless.i`. So it already depends on the context.

isuffix:
I could be open to allowing `i.dotless` to act like a field access on `i`. It seems like there are multiple reasonable uses and it makes some sense given the one-letter modifiers on other symbols like `arrow.r`. This would break the general rule that single letters don't act like identifiers, but that is a lexical rule, and adding a dot and another identifier does seem to create a new lexical context where a field access could be reasonably intended.

However, I am concerned about creating errors in expressions like `i.x` where `x` isn't a field of `i`. Currently Typst lets you be quite sloppy with dots and single characters, and I find it to be reasonably ergonomic that I don't have to consider errors for `x.field` or `x.y`. A common example where those might come up would be the `forall` and `exists` quantifiers, where `$forall a.a=a$` would now error.

Another example would be the lambda calculus, where expressions like `$lambda x. lambda y. x y$` are quite common, and if a user isn't careful with their spaces, either of `x.lambda` or `y.x` would start to produce an error with this change. I did a bunch of lambda calculus with Typst for some assignments in university, and (with some regular expressions) I can find over a dozen examples of me writing both the `x.lambda` and `y.x` forms, so this is a strong concern for me.

And while we could recognize `x.lambda` as not being a valid field access and render it as `x . lambda` at runtime, this would lead to asking whether `pi.lambda` might fallback to rendering as `pi . lambda`, which I find absurd, so I would not consider that option.

Overall, I agree that this could be a better interface for these characters, but I don't think allowing this syntax would be worth the ergonomic price in strictness around single characters and dots.

MDLC01:
Thanks for the detailed answer.

Since I am the only one in favor of supporting `i.dotless`, I think the next best option is to use a submodule rather than a symbol. I agree with this:

> I can see that the current state can seem no big deal but is there a downside to the proposed change (other than breaking the code of someone that writes dotless instead of dotless.i)? It does seem conceptually more correct...

CHANGELOG.md
```diff
@@ -127,6 +127,9 @@ These previously deprecated items were removed:
 - `shin`
 - `planck.reduce`
 
+The `dotless` symbol was converted to a module,
+meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
```
MDLC01:
```suggestion
- The `dotless` symbol was converted to a module,
  meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
```

Additionally, I think this should be in the General Changes section.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### General changes
+
+- The `dotless` symbol was converted to a module,
+  meaning you can no longer use `dotless` as a shorthand for `dotless.i`.
```
MDLC01:
```suggestion
- The `dotless` symbol was converted to a module,
  meaning you can no longer use `dotless` as a shorthand for `dotless.i`. **(Breaking change)**
```

Sorry for the two consecutive changes at the same place.

MDLC01 (changes requested):
I don't see the absence of a deprecation period as an issue that should block us from making this change. Probably very few users use bare `dotless` anyway.

T0mstone:
I can drop ddddfa2, but it's such a minor change that I think doing it along the way here is fine.

MDLC01 (approved):

sijow (approved):

MDLC01:
We are trying a new way to keep track of changes by maintaining the changelog in a separate shared document (https://typst.app/project/r2ChBGT9XWBoEhAKGM6icb). Feel free to drop the update to the changelog from this PR to prevent any further conflict.

Also, this will be missing a third approval to be merged after the conflict is resolved.

PR #160, opened 2026-04-29
Add `eq.dot`

### LINKED ISSUE #159: Add `eq.dot` for ≐
There are `eq.dots` and `eq.quest`, would be nice to have `eq.dot` for `\u{2250}` (≐).

NamorNiradnug:
Closes #159

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
   - `lt.closed.eq.not`: ⋬
   - `lt.closed.not`: ⋪
   - `eq.dot`: ≐
+  - `equiv.dot`: ⩧
 
 - Miscellaneous technical
   - `bowtie.stroked`: ⋈
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
   - `lt.closed.eq`: ⊴
   - `lt.closed.eq.not`: ⋬
   - `lt.closed.not`: ⋪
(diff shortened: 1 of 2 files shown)


MDLC01 (approved):

NamorNiradnug:
It seems appropriate to add `equiv.dot` (⩧) in this PR by the way. Should it be added?

dccsillag (approved):

dccsillag:
> It seems appropriate to add equiv.dot (⩧) in this PR by the way. Should it be added?

Sounds reasonable to me. I must admit I have never seen would-be `equiv.dot` used before, but it would be consistent with `eq.dot`.

Enivex:
There's plenty of other symbols with dots above, e.g. ⩑ ⨰ ⫃ ⩭

Unfortunately the situation is already inconsistent, since (among other symbols) `and.dot` is currently ⟑ instead of ⩑.

I really feel like there needs to be a more holistic solution here.

MDLC01:
I think for now let's only add ≐ because it is widely used and unlikely to end up with a different name in the end. We can add ⩧ later, when we decide how to handle all symbols with decorations.

dccsillag:
agreed

MDLC01:
Just as a heads up, I will merge this in about five days unless someone objects.

dccsillag:
Sounds good to me.

PR #164, opened 2026-05-10
Add keyboard labels symbols

zacharie-moughanim:
This PR aims at naming symbols that are often used as keyboard labels.
This regroups within a submodule - as was suggested in [this PR](https://github.com/typst/codex/pull/153) - symbols coming from across several unicode blocks : Arrows, Supplemental Arrows - B, Control Pictures, Miscellaneous Technical, Miscellaneous Symbols and Arrows, Miscellaneous Symbols and Pictographs.
There are probably some other I've missed, if you know some, I'll add them.
I've tried to add a large proportion of symbols that _may_ be in the submodule in the end. Some I've considered and did not add, some I'm not convinced should stay but I put anyway to open the discussion. See [this table](https://typst.app/project/roIJqAoI7n5XT088yx7pyW) to see all the symbols I considered.

Here are some remarks about the choices of names that I think needs to be discussed : 
- Some arrows (`↸`/`⭶`, `⇲`/`⭸`, `↹`/`⭾`, `⇤`/`⭰`, ...) both exists as simple arrows, and yet a very similar arrow exist with the main purpose of designating a keyboard label in the Unicode chart. I chose to put arrows resembling-but-whose-main-purpose-is-not-being-a-keyboard-label as `alt`. However this becomes weird for three tab symbols: `↹`/`⭾`/`⭿`; the third one is a  vertical tab, and is rotation of the second one, being itself an arrow but not primarily a keyboard label, while the actual designated symbol for keyboard is `↹` and has no vertical counterpart. For now, `tab.alt.v` is set to the `⭿`, so `tab.v` falls back to it.
- I named four variants of `capslock`. Only one of them is not designated as capslock in the Unicode chart (`⮸`), but I put it for its resemblance, and the fact that it's a combination of variation on other symbols designated as capslock.
- I used the variant name `struck` for some capslock to designate a horizontal bar inside the stroked arrow e.g. `⇫`/`⇬` although it does not correspond to what is called `struck` for other arrows. I'm looking for a better name for this variant.
- There are 3 symbols designating a removal of inputs : `␥`, `⌦`, and `⌫`, respectively named `del`, `erase.r` and `erase.l`. The first name is for consistence with the control pictures block (it's an alternate symbol for a symbol named del in the control submodule here), however there is also `erase.l` that I think is used to designate the same key (namely, the one deleting characters to the left of the cursor, so the naming feels weird here.
- The symbol for the option key  `⌥` is also sometimes used for the alt key (description of the Unicode chart). There is another symbol for the alt key: `⎇` and I don't know if it's a good idea to put the first one a second time in the module as an alternate symbol for the `alt` name (Unrelated but `alt.alt` would be a bit weird and I have no idea what the name could be otherwise).
- Sometimes regrouped under "Keyboard symbols" on Unicode charts - along with symbols used as keyboard labels - are symbols representing a keyboard. I named them but I don't think they should be in this submodule (plus they are currently named `keyboard.keyboard` which, in my opinion, feels off like `alt.alt`).
- I didn't put any simple arrows that could stand for the arrows at the bottom right of the keyboard because I did not find a unicode symbol mentioning these keys. I'm curious if you have any idea what arrow to pick here, or if we should pick one at all.
- The arrow `↵` (from the Arrow block) is described as "may represent return or newline on keyboards" in the Unicode chart. Since there are other symbols for both these keys, I put it with both names `return.alt` and `newline.alt`.
- I had difficulty drawing the line between what should and should not be in this submodule. Among symbols I'm not sure should be here, and are in the submodule right now are symbols to adjust sound. Some I think should stay like speakers symbols (used for increase/mute/... the sound on a keyboard). Nonetheless, I feel like some, like fast-forward, are more MP3's or DVD player's button labels rather than keyboard's. Among those I did _not_ put are lock 🔒 and bell 🔔. (All considered symbols, present or not in the submodule, are in the table linked at the beginning of this message).
- As was discussed in [the PR about control pictures](https://github.com/typst/codex/pull/153), I removed the `spacebar` for `␣` and moved this symbol in the submodule, whose full name is now `keyboard.space`.  Plus, I think it should be noted that a similar symbol called SHOULDERED OPEN BOX `⍽` (the previous one is OPEN BOX) exists, although I did not name it here since I don't know if it's used as a keyboard label, tell me if it is used as such.


Enivex:
I'm _really_ not a fan of adding generic symbols with such specific names. Only symbols clearly intended to represent keyboard keys should be introduced this way.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
   del ␡ // DELete
   nl ␤ // New Line
 }
+
+// Characters commonly used as keyboards labels
+keyboard {
+  linefeed ↴
```
MDLC01: Is this symbol present on actual keyboards?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  home ↸
+    .alt ⭶
```
MDLC01: I don't like that we have two subtly different symbols for the same thing. How did you choose which one to be the `.alt` variant? Also, why not simply ↖ as suggested by [French Wikipédia](https://fr.wikipedia.org/wiki/Touche_de_d%C3%A9but)?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+    .corner ⇱
+    .phone ⌂
```
MDLC01: There could be many more variations (like ↖ which I noted previously).

In general, I feel like if a semantic name doesn't fit a single symbol (with perhaps a few variants, like `phi` and `phi.alt`), it would be better to simply abandon the semantic name and use descriptive names for all variants instead.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  end ⇲
+    .alt ⭸
```
MDLC01: Why is the default for this one not the same as for `home`?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+    .alt ⭾
+    .alt.v ⭿
+    .l ⇤
+    .l.alt ⭰
+    .r ⇥
+    .r.alt ⭲
+    .t ⤒
+    .t.alt ⭱
+    .b ⤓
+    .b.alt ⭳
```
MDLC01: Do you have evidence for any variant other than `tab` and `tab.alt` being used in practice?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  pageup ⇞
+  pagedown ⇟
+  shift ⇧
+  capslock ⇪
+    .pedestal ⇫
+    .pedestal.struck ⇬
+    .struck ⮸
```
MDLC01: Same comment as https://github.com/typst/codex/pull/164/changes#r3215401498.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  numlock ⇭
```
MDLC01: Do you have evidence of this symbol being used on actual contemporary keyboards? I personally don't associate any symbol with the Num Lock key.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  scroll ⇳
```
MDLC01: Is this a real keyboard key?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  space ␣
+  del ␥
```
MDLC01: Do you have evidence of contemporary real world keyboards that use this symbols?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  erase
+    .r ⌦
+    .l ⌫
+  cmd ⌘
+  enter ⌤
+    .alt ⎆
```
MDLC01: I don't think most people associate the Enter key with any of those symbols.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  option ⌥
+  clear ⌧
```
MDLC01, line 1328: Is this a real key on contemporary keyboards?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  keyboard ⌨\vs{text}
+    .mouse 🖦
+    .wired 🖮
```
MDLC01: I don't think this fits in a submodule dedicated to keyboard key symbols.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  insert ⎀
+  underline
+    .continuous ⎁
+    .discontinuous ⎂
+  emph ⎃
+  center ⎅
```
MDLC01: Do you have evidence of contemporary keyboards that have dedicated keys for those functions?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  alt ⎇
+  control ⎈
```
MDLC01: Do you have evidence of contemporary keyboards that use this symbol?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  break ⎊
```
MDLC01: What key is that?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  esc ⎋
+    .alt ⮹
+  undo ⎌
```
MDLC01: Same as https://github.com/typst/codex/pull/164/changes#r3215411291.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  decimalsep ⎖
```
MDLC01: What key is that?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  prevpage ⎗
+  nextpage ⎘
```
MDLC01: This feels redundant with `pageup` and `pagedown`.

zacharie-moughanim:
You can see the table linked in the initial message to get all informations but quick recap :
Here are symbols not explicitly mentioned as key symbols :
- ⌂ HOUSE in Miscellaneous technical > Miscellaneous technical which, as said in the document, I don't know if it's actually used as a keyboard label, but other key symbols are used to denote "HOME KEY", I put it to open the discussion but I'm not convinced it should stay.
- ⌨ and derivatives which, even if it is in a group with key labels, represents a keyboard rather than being a key, so I don't think it belongs here.
- User interface symbols : none of U+23E9 - U+23FA should in my opinion be there and only a few among U+1F500-U+1F5AE, mainly speakers and brightness frequently used as keyboard labels although it is more generally a UI symbol for brightness and sound management.
- Power symbols U+23FB - U+23FE and U+2B58, same remark as for brightness and speakers.
- Lastly, there are generic arrows (from Arrows and Supplement-B Arrows blocks) _described as key labels_, but I'm not convinced whether to keep them or not. For instance : ⭰, see the initial message for a thorough explanation.

Feel free to tell me what you think, which among those (maybe all of them or more) you wish to see removed from this PR.

src/modules/sym.txt
MDLC01: This one specifically is not used as a common keyboard key symbol so I don't think it belongs here.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  pageup ⇞
+  pagedown ⇟
+  shift ⇧
+  capslock ⇪
+    .pedestal ⇫
+    .pedestal.struck ⇬
+    .struck ⮸
+  numlock ⇭
+  scroll ⇳
+  space ␣
```
MDLC01, line 1322: Is this even present on actual keyboards? I feel like most keyboards simply do not have any symbol on the space bar.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  del ␥
+  erase
+    .r ⌦
```
MDLC01: Is this present on any real world keyboard?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  control ⎈
+  break ⎊
+  esc ⎋
+    .alt ⮹
+  undo ⎌
+  decimalsep ⎖
+  prevpage ⎗
+  nextpage ⎘
+  printscreen ⎙
+  clearscreen ⎚
```
MDLC01: Are those symbols used on any real world contemporary keyboard?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  return ⏎
+    .alt ↵
+    .l ⮐
+    .r ⮑
```
MDLC01: I think those can be excluded for now.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  newline
+    .alt ↵
+    .l ⮒
+    .r ⮓
```
MDLC01: Is there any contemporary keyboard that has a separate key for newline that is not just the Enter key?

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  decrease ⏷
+  pause ⏸\vs{text}
+    .alt ⎉
+  stop ⏹\vs{text}
+  record ⏺\vs{text}
+  shuffle 🔀
+  repeat 🔁
+    .once 🔂
+  reload 🔃
+    .alt 🔄
```
MDLC01: I think let's exclude those from this PR. they can be added in a separate PR so that this one can focus on actual keyboard symbols.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  // Adjust sound and brightness
+  brightness
+    .decrease 🔅
+    .increase 🔆
+  speaker 🔈\vs{text}
+    .mute 🔇
+    .lower 🔉
+    .higher 🔊
```
MDLC01: I feel like those symbols can be interpreted in a more general way than just keyboard symbols. Let's add them in a separate PR.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  // Power symbols
+  power ⏻
+    .on ⏽
+    .off ⭘
+    .on.off ⏼
+    .sleep ⏾
```
MDLC01: Those are redundant with `sym.power.standby`, `sym.power.on`, `sym.power.off`, `sym.power.on.off` and `sym.power.sleep`. also they are not keyboard symbols, so I don't think they belong here.

src/modules/sym.txt
```diff
@@ -1324,3 +1320,112 @@ control {
...
+  keyboard ⌨\vs{text}
+    .mouse 🖦
+    .wired 🖮
+  insert ⎀
+  underline
+    .continuous ⎁
+    .discontinuous ⎂
+  emph ⎃
+  center ⎅
+  alt ⎇
```
MDLC01, line 1329: Do you have evidence of contemporary keyboards that use this symbol?

MDLC01 (commented):
In general, I feel like a lot of symbols can be removed (for most of them, I left a comment asking for evidence of the symbols being used on actual contemporary people, i.e., being what users expect). Some can be added in a separate PR so that they can be discussed separately as a set.

To answer the various points you made in your OP:
1. Let's think when we have a smaller set of symbols to consider.
2. Addressed as a separate comment.
3. I agree that we should not use `.struck` for that.
4. Addressed as separate comments.
5. Let's keep a single name for each symbol, especially when the alternate name would be as an `.alt` of a different symbol.
6. I also agree they shouldn't be in this submodule.
7. I don't think there is a point giving hyper-specific semantic names to symbols that already have descriptive names in Codex. Also, various arrows fit this semantic, so let's let the end user choose the ones they want instead of having opinionated ones.
8. I don't like having different words for the same thing. I understand the problem: when multiple symbols have essentially the same semantic, there is a naming conflict. But I think the solution here is to simply choose one and give descriptive names to the others (in this case, ↵ currently does not have a semantic name but let's keep that for a separate PR adding more arrows[^1]).
9. Addressed as separate comments.
10. Addressed as a separate comment.

I did not go into a lot of details when reviewing this PR as I first want to reduce its scope to make it more manageable. Feel free to point out any oversight.

[^1]: See also the huge Arrows section in the [Symbol List for `sym`](https://typst.app/project/riXtMSim5zLCo7DWngIFbT) document.

src/modules/sym.txt
zacharie-moughanim: As explained in the linked document, the default one is in a group called "Keyboard symbols and circle arrows" while the other is in a random group (here, "triangle-headed arrows"), but both mention the "home key" and are cross-referenced. So they clearly meant to describe the same key, but the first one is primarily meant to be a key label.

src/modules/sym.txt
zacharie-moughanim: The difference is that ↖ is not described as "home" in the unicode chart. But if you want we can remove home altogether.

src/modules/sym.txt
MDLC01: Then I think let's only keep U+21B8 and remove the `.alt` variant. The way I understand it, U+2B76, as a specifically triangle-headed arrow, is a more specific version of U+21B8, a generic arrow.

src/modules/sym.txt
MDLC01: I think that would be better indeed. If the need for a `keyboard.home` symbol arises in the future, we can always add it at that point.

src/modules/sym.txt
T0mstone: My Apple keyboard has U+21E5 on it.

src/modules/sym.txt
T0mstone: My Apple keyboard has it (U+2327, tho the X slightly pokes out). But I have no clue what it was meant to represent, since I'm using this keyboard on Linux, where it maps to num lock.

src/modules/sym.txt
T0mstone: Mine uses a filled version of U+2326 with a white cross for delete, next to "delete" written out on the key.

src/modules/sym.txt
zacharie-moughanim: I sometimes find keyboards with a key for 'Enter' and a key for 'Return' (e.g. [this one](https://www.startpage.com/av/proxy-image?piurl=https%3A%2F%2Fpisces.bbystatic.com%2Fimage2%2FBestBuy_US%2Fimages%2Fproducts%2F6276%2F6276710_sd.jpg&sp=1782581015T233ba5931dbbf477121a8ce909131585c8211166a98c86edb65dbe0de2a4d34d)) often when the keyboard has a numeric pad.
Although I think I never saw the symbols described in the unicode chart as 'Enter' as an enter key.

I propose to keep either ('enter' and 'return') or ('newline' and 'return') or keep only one name, in which case I would favor 'enter', tell me what you think.
Plus, would you rather keep all these symbols but merge them under a single name or remove some symbols ?

src/modules/sym.txt
zacharie-moughanim: [This keyboard](https://www.ldlc.com/fiche/PB00556046.html) has U+2327.

src/modules/sym.txt
zacharie-moughanim: [This keyboard.](https://www.asus.com/accessories/keyboards/tuf-gaming/tuf-gaming-k3-gen-ii/) has U+21EA.
My lenovo keyboard (below) has U+21EB.
Image

src/modules/sym.txt
zacharie-moughanim: [This keyboard](https://www.ldlc.com/fiche/PB00592746.html) has U+2398. Since apparently both are used, I'll merge pageup/down with prev/nextpage.

src/modules/sym.txt
zacharie-moughanim: For now I'll remove the other two.

src/modules/sym.txt
zacharie-moughanim: [This keyboard](https://www.ldlc.com/fiche/PB00610154.html) has U+2326.

src/modules/sym.txt
zacharie-moughanim: Image
It appears on the display of 'Keyboard layout' in the settings of my ubuntu 24.04, though I guess this means it is used on actual keyboards, I didn't find any.

src/modules/sym.txt
zacharie-moughanim: Most keyboards I've seen simply have a straight line for spacebar, I found that some [phone keyboard on android](https://www.lifewire.com/thmb/NITYBo8e6ZlfTa4bRNMuC_6xn3M=/1500x0/filters:no_upscale():max_bytes(150000):strip_icc()/Screenshot1-895a0204e8e7429ca375364d38dbf258.jpg) use this symbol.

src/modules/sym.txt
zacharie-moughanim: Adds a symbol representing a separator between the integer and fractional part of a number.
For instance, with a french keyboard layout, the sequence of keystrokes [0] [decimalseparator] [5] is print out 0,5 while it is 0.5 with an english layout.
Such a key is sometimes found near numeric pad on large keyboard, although the label is often just a dot or a comma.

src/modules/sym.txt
zacharie-moughanim: Some keyboard have a symbol for the numlock key (e.g. [this one](https://www.ldlc.com/fiche/PB00556046.html), [this one](https://www.ldlc.com/fiche/PB00279755.html) and [this one](https://www.ldlc.com/fiche/PB00103723.html)), but I didn't found this particular symbol. I'll remove it from this PR for now.

src/modules/sym.txt
zacharie-moughanim: I'll remove the others for now.

src/modules/sym.txt
zacharie-moughanim: I didn't find any, I'll remove it for now.

src/modules/sym.txt
zacharie-moughanim: This symbol, framed within a rectangle, appears on [this keyboard](https://static0.howtogeekimages.com/wordpress/wp-content/uploads/2022/07/1-scroll-lock-key-on-keyboard.jpg).

src/modules/sym.txt
zacharie-moughanim: I didn't find any, I'll remove this symbol.
Do you think U+2326 should be renamed `del` or should it remain `erase.r` ?

src/modules/sym.txt
zacharie-moughanim: According to [wikipedia](https://en.wikipedia.org/wiki/Insert_key#Symbol), even though the insert symbol is the ISO symbol for the insert key, it is implied that it's not used that way.
I didn't find this one nor the other on keyboards. I'll remove them.

src/modules/sym.txt
zacharie-moughanim: I didn't find any, I'll remove it.

src/modules/sym.txt
zacharie-moughanim: The [break/pause key](https://en.wikipedia.org/wiki/Break_key) present on large keyboards, I've seen it written out but not this symbol, I'll remove it.

src/modules/sym.txt
zacharie-moughanim: I didn't find any, I'll remove it.

src/modules/sym.txt
zacharie-moughanim: I've frequently seen symbols for the print screen key (e.g. [this one](https://www.ldlc.com/fiche/PB00461747.html)), but not this one, tell me if you'd rather I remove it from this PR.
I'll remove clearscreen, I didn't find this symbol for this key.

src/modules/sym.txt
zacharie-moughanim: I'll remove these symbols.

src/modules/sym.txt
zacharie-moughanim: There are two symbols described as 'end' : U+2B78 (regrouped with U+2B76 representing home, which has been removed from this PR) and U+21F2 (regrouped with U+21F1 representing home, which has also been removed) while no symbol for 'end' is directly near the only symbol for home we kept (U+21B8).

src/modules/sym.txt
zacharie-moughanim: I did as you requested.

src/modules/sym.txt
zacharie-moughanim: I removed them.

src/modules/sym.txt
zacharie-moughanim: It has been removed.

src/modules/sym.txt
zacharie-moughanim: I removed them.

src/modules/sym.txt
MDLC01: This was for the choice of the arrow variant, but as you suggested in https://github.com/typst/codex/pull/164#discussion_r3222378176, I agree it is better to remove `home` altogether.

src/modules/sym.txt
MDLC01: I think we should probably remove this one as well if we remove `home` (or even if we keep `home` but can't make both consistent).

src/modules/sym.txt
MDLC01: I would remove the `.alt` variants as well, for the same reason as https://github.com/typst/codex/pull/164#discussion_r3222779836.

src/modules/sym.txt
MDLC01: How did you choose the default between the variant with a pedestal and the variant without a pedestal?

src/modules/sym.txt
MDLC01: Then, this would be more specifically for scroll lock. I think we should leave this symbol out for now, especially because the keyboard you linked uses a different version, and frames it in a rectangle-like shape as you noted.

src/modules/sym.txt
MDLC01: I think `erase.r` is fine

src/modules/sym.txt
MDLC01: I don't know what this is used for, but this symbol is indeed used for the Clear key on Apple Keyboards.[^1] I think we can keep it.

[^1]: https://en.wikipedia.org/w/index.php?title=Apple_keyboards&oldid=1362577651#mwVg

src/modules/sym.txt
MDLC01: Unless this is a widely recognized label, I would remove it.

src/modules/sym.txt
MDLC01: How did you choose which variant to be the `.alt`?

src/modules/sym.txt
MDLC01: I think this is fine as is because I suspect this is a widely recognized symbol for spaces

src/modules/sym.txt
MDLC01: If actual keyboards use different symbols, let's remove `printscreen`.

src/modules/sym.txt
MDLC01: I think we should have at most a single name. I am also in favor of `enter`.

Regarding the value, I would be in favor of having U+21B5 as the main value and U+23CE as the `.alt`, but this is mostly due to personal preference.

src/modules/sym.txt
```diff
@@ -1250,9 +1250,6 @@ errorbar
   .circle.stroked ⧲
   .circle.filled ⧳
 
-// Keyboard labels.
-spacebar ␣
-
```
MDLC01, line 1255: Could you please deprecate the symbol instead of removing it? This can be done as follows:

```
@deprecated: `spacebar` is deprecated, use `keyboard.space` instead
spacebar ␣
```

MDLC01:
We are trying a new way to keep track of changes by maintaining the changelog in a separate shared document (https://typst.app/project/r2ChBGT9XWBoEhAKGM6icb). Feel free to drop the changelog update from this PR to prevent any further conflict.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: It was the most common among actual keyboards I saw when searching symbols for this PR.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I first chose the most common pair of symbols among actual keyboards, I think it's U+2B7B/U+2B7D. Then, I chose the canonical names used in the unicode chart for this symbol.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

CHANGELOG.md
```diff
@@ -103,9 +103,6 @@ These previously deprecated items were removed:
 - Currency
   - `riyal`: ⃁
 
-- Keyboard labels
-  - `spacebar`: ␣
-
```
MDLC01, line 108: Please remove this as well so that the changelog contains no change at all from this PR.

zacharie-moughanim: It's from [a previous PR](https://github.com/typst/codex/pull/153/changes), do you want me to remove it anyway ?

MDLC01: Actually, I had not realized that this was modifying the changelog for a released version. Then, we definitely do not want to have this change. The goal of each changelog is to document the changes that have landed in a specific version, so it is expected that older changelogs may mention symbols that no longer exist in the most recent versions.

src/modules/sym.txt
```diff
@@ -1307,3 +1304,36 @@ control {
...
+    .alt ⎗
+  pagedown ⇟
+    .alt ⎘
+  shift ⇧
+  capslock ⇪
+    .pedestal ⇫
+  space ␣
+  erase
+    .r ⌦
+    .l ⌫
```
```suggestion
    .l ⌫
    .r ⌦
```

Users probably expect bare `erase` to fall back to `erase.l` rather than `erase.r`.

src/modules/sym.txt
```diff
@@ -1307,3 +1304,36 @@ control {
...
+  cmd ⌘
+  option ⌥
+  clear ⌧
+  alt ⎇
+  esc ⎋
+    .alt ⮹
+  enter ↵
+    .alt ⏎
```
MDLC01: Thinking about it more, I am now tempted to go against what I suggested in https://github.com/typst/codex/pull/164#discussion_r3528125555 and switch the main symbol and the alternate version, because U+23CE ⏎ is intended primarily as a keyboard symbol, whereas U+21B5 ↵ is more general purpose, and therefore may be given a more generic name in Codex in the future.

zacharie-moughanim: Done.

MDLC01:
There are still some review comments to address. I believe this is an exhaustive list:
- https://github.com/typst/codex/pull/164#discussion_r3527734311
- https://github.com/typst/codex/pull/164#discussion_r3527746781
- https://github.com/typst/codex/pull/164#discussion_r3527761951
- https://github.com/typst/codex/pull/164#discussion_r3553604401

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

src/modules/sym.txt
zacharie-moughanim: I did as you suggested.

CHANGELOG.md
zacharie-moughanim: I did as you suggested.

MDLC01 (approved):

src/modules/sym.txt
```diff
@@ -1307,3 +1304,31 @@ control {
...
+  space ␣
+  erase
+    .l ⌫
+    .r ⌦
+  cmd ⌘
+  option ⌥
+  clear ⌧
+  alt ⎇
+  esc ⎋
+    .alt ⮹
```
T0mstone: Has anyone seen these on an actual keyboard?

sijow: The symbol ⎋ is used on some Mac keyboards and in macOS, see for example https://commons.wikimedia.org/wiki/File:Macintosh_escape_key.jpg and https://support.apple.com/guide/mac-help/what-are-those-symbols-shown-in-menus-cpmh0011/mac

I couldn't find any use of the alternate symbol.

zacharie-moughanim: I searched for a bit and found no evidence of `esc.alt` used as a keyboard symbol except the Unicode chart. It's therefore probably better to use a descriptive name for this symbol, I removed it from this PR.

sijow (approved):

T0mstone (approved):

T0mstone:
Disregard the approval message, that was left over from when I was drafting my previous comment😅

PR #165, opened 2026-05-12
Bump transitive `rustls-webpki` dependency

T0mstone:
...so that I can stop getting the annoying dependabot alert E-mails.
(I also considered changing my notif settings, but since there's apparently a vulnerability, why not stop people from running vulnerable code when running the tests that fetch something from the web)

We can then later remove the extra dependency line again after `ureq` updates its `rustls` dependency and we update to that newer `ureq` version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ keywords = ["unicode", "symbols"]
 default = ["numeral-systems", "styling"]
 numeral-systems = ["dep:chinese-number"]
 styling = []
-_test-unicode-conformance = ["ureq"]
+_test-unicode-conformance = ["dep:ureq", "dep:rustls-webpki"]
 
 [dependencies]
 chinese-number = { version = "0.7.7", default-features = false, features = ["number-to-chinese"], optional = true }
@@ -25,3 +25,5 @@ siphasher = "1.0.2"
 
 [build-dependencies]
 ureq = { version = "3.0.12", optional = true }
+# Security fix. Can be removed again with a ureq update that transitively depends on at least this version.
+rustls-webpki = { version = "0.103.13", optional = true }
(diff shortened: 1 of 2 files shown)


MDLC01:
I don't understand why we need rustls-webpki even though it's not used in the code

T0mstone:
It's a transitive dependency (ureq depends on rustls, which depends on rustls-webpki). Adding an it as an explicit dependency tells Cargo's version resolver to use that version as the minimum, which is higher than what it would otherwise have chosen.

Cargo.toml
```diff
@@ -25,3 +25,4 @@ siphasher = "1.0.2"
 
 [build-dependencies]
 ureq = { version = "3.0.12", optional = true }
+rustls-webpki = { version = "0.103.10", optional = true }
```
MDLC01: The patched version in the security alert is 0.103.13. Why did you choose 0.103.10 here? It seems to resolve to 0.103.13 in Cargo.lock, but why not explicitly ask for it here?

T0mstone: Oops, my bad, I'll change it.

mkorje (approved):

MDLC01 (approved):

MDLC01:
Let's merge this before we get any further Dependabot e-mails.

laurmaedje:
> It's a transitive dependency (ureq depends on rustls, which depends on rustls-webpki). Adding an it as an explicit dependency tells Cargo's version resolver to use that version as the minimum, which is higher than what it would otherwise have chosen.

It's not really necessary to add it. You can just run `cargo update -p rustls-webpki` to bump the transitive dep in `Cargo.lock`. This will not affect library consumers, but those would be expected to update their own lock files.

MDLC01:
Let's remove it next time we update `Cargo.toml` then, as it's not causing any issue right now.

Issue #166, opened 2026-05-13
Make sure all emoji are fully qualified


MDLC01:
we may want to make sure all emoji in `emoji` are fully qualified. This is more complex than just having an emoji presentation selector at the end. Sometimes it is in the middle, such as for `2764 FE0F 200D 1F525` ❤️‍🔥.

Data at https://www.unicode.org/Public/17.0.0/emoji/emoji-test.txt. More succinct data which seems to only contain `fully-qualified` and `component` entries: https://unicode.org/emoji/charts/emoji-ordering.txt.

~~Currently, two emoji are not even valid: `ast` should be `002A FE0F 20E3` and `hash` should be `0023 FE0F 20E3`. They are both missing U+20E3.~~ See: https://github.com/typst/codex/pull/167#issuecomment-4447816396.

I am not sure whether we actually want to do that. I am opening this issue now to make sure I don't forget to look into it later.

MDLC01:
Note that the documents linked above are not exhaustive lists of all emoji (see https://github.com/typst/codex/pull/167).

PR #167, opened 2026-05-13
Fix two emoji

MDLC01:
Two emoji are missing a combining enclosing keycap component. With this fix, all emoji in `emoji` should be part of https://www.unicode.org/Public/17.0.0/emoji/emoji-test.txt, and more specifically https://unicode.org/emoji/charts/emoji-ordering.txt, which seems to only include fully-qualified forms and components.

Related: https://github.com/typst/codex/issues/166.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,8 +242,12 @@ mod test {
         let sequences = get_valid_presentation_sequences();
         assert!(
             are_all_variants_valid(ROOT, |c| {
+                // FIXME: Some emoji consist of many characters and contain an
+                // emoji presentation sequence in the middle, but are not listed
+                // in emoji-variation-sequences.txt.
+                // See: https://github.com/typst/codex/issues/166.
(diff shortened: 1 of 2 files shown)


T0mstone:
I think both are valid. https://www.unicode.org/Public/17.0.0/ucd/emoji/emoji-data.txt contains # and *, both in the basic Emoji category, and further down as an Emoji_Component.

As further evidence, here's a screenshot of a font rendering # and * with text presentation, emoji presentation, and emoji presentation + combining keycap
screenshot_1

So I think these should be new variants, perhaps `hash.keycap` and `ast.keycap`.

MDLC01:
So I did a bit of research, and I think there is no exhaustive list of emoji. Meaning the absence of bare * and # from some data files indeed does not indicate that they aren't emoji.

I will close this PR and left some additional comments at https://github.com/typst/codex/issues/166#issuecomment-4450153215 and https://github.com/typst/typst/pull/8282#issuecomment-4450154608.

PR #169, opened 2026-05-27
Add Abjad letters

### LINKED ISSUE #168: Arabic alphabetic numbering for lists and headings.
### Description

I need to add alphabetic numbering for this small example, using arabic letters
 أ,ب,ج,....  `numbering: "1.a)"` give latin alphabet in second level.

```typst
#set text(lang: "ar", font: "Amiri")

#set heading(numbering: "1.a)")
#set enum(numbering: "1.a)")

= عنوان
== عنوان فرعي

+ مثال على قائمة 
+ مثال على قائمة 
   + مثال على قائمة 
   + مثال على قائمة 

```

Image

### Use Case

The problem arises for all languages, users can insert numbering in their own alphabets.

seloumi:
Add Arabic abjad numbering #168

### DIFF
--- a/src/numeral_systems.rs
+++ b/src/numeral_systems.rs
@@ -1467,7 +1467,7 @@ mod tests {
                 NamedNumeralSystem::KoreanJamo => 65477685939649764827530478995838083425, // 21
                 NamedNumeralSystem::KoreanSyllable => 24217153056183571894327643661698510954,
                 NamedNumeralSystem::EasternArabic => 277754701051910363703826860323053920831,
-                NamedNumeralSystem::ArabicAbjad => 0, 
+                NamedNumeralSystem::ArabicAbjad => 332217778941065507236060222268877272547, 
                 NamedNumeralSystem::Persian => 6232158096065129450489636457808686806,
                 NamedNumeralSystem::Devanagari => 327133969362282954753636774557232534052,
                 NamedNumeralSystem::Tibetan => 87580519645280744681237273097105390953,
--- a/src/numeral_systems.rs
+++ b/src/numeral_systems.rs
@@ -408,6 +408,33 @@ declare_named! {
         /// > <span dir="auto">٨</span>, <span dir="auto">٩</span>,
         /// > <span dir="auto">١٠</span>, <span dir="auto">١١</span>
         EasternArabic = "arabic.eastern" ("١"),
+        
+        /// Abjad order
+        /// [Arabic letters](https://en.wikipedia.org/wiki/Abjad).
+        ///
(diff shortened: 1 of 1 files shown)


src/numeral_systems.rs
```diff
@@ -1435,6 +1467,7 @@ mod tests {
                 NamedNumeralSystem::KoreanJamo => 65477685939649764827530478995838083425, // 21
                 NamedNumeralSystem::KoreanSyllable => 24217153056183571894327643661698510954,
                 NamedNumeralSystem::EasternArabic => 277754701051910363703826860323053920831,
+                NamedNumeralSystem::ArabicAbjad => , // ?
```
MDLC01: To figure out what value to put here, first try an arbitrary value like `0`. It's not gonna work, but the error message in GitHub CI will tell you the right value.

MDLC01: So the value is `332217778941065507236060222268877272547`

src/numeral_systems.rs
```diff
@@ -821,6 +848,11 @@ impl NamedNumeralSystem {
             Self::EasternArabic => NumeralSystem::Positional(&[
                 '٠', '١', '٢', '٣', '٤', '٥', '٦', '٧', '٨', '٩',
             ]),
+            
+            Self::ArabicAbjad => NumeralSystem::Positional(&[
+                'أ', 'ب', 'ج', 'د', 'ه', 'و', 'ز', 'ح', 'ط', 'ي', 'ك', 'ل', 'م', 'ن', 
+                'س', 'ع', 'ف', 'ص', 'ق', 'ر', 'ش', 'ت', 'ث', 'خ', 'ذ', 'ض', 'ظ', 'غ',
+            ]),
```
MDLC01, line 855: If I understand properly, this is a system where the first letter has value one, then all the way up to the last letter of the alphabet, and then you start using multiple letters? If this is the case, you should use `Bijective` instead of `Positional`. Otherwise, could you please elaborate on how this numeral system works?

seloumi: Users will mostly not need all this numbering; the first ten are sufficient.

MDLC01: The most important reason why we want to use `Bijective` is that the first letter should have the value one instead of zero. For example, when using numbered headings in Typst, if using `Positional`, the first letter would always be skipped because the first heading has the number one.

seloumi:
I will close this and send a new request

MDLC01:
Don't worry about having a bad-looking commit history on your pull requests. It's all gonna get squashed in the end anyway :)

PR #170, opened 2026-05-28
Add Arabic abjad numerals

seloumi:
Arabic abjad numeral for alphabetic numbering #168.

PR #171, opened 2026-05-29
Use strings instead of chars in `NumeralSystem`

MDLC01:
This is required for https://github.com/typst/codex/pull/170 because the number five is represented by the string `"ه‍"`, which consists of two codepoints: <U+0647, U+200D>.

I changed every numeral system that previously used `char` to use `&str` instead.


src/numeral_systems.rs
```diff
@@ -1229,7 +1229,8 @@ mod tests {
     /// Makes sure fixed systems are implemented properly.
     #[test]
     fn test_fixed_systems() {
-        let symbols = ('a'..='z').collect::<Vec<_>>();
+        let symbols = ('a'..='z').map(String::from).collect::<Vec<_>>();
+        let symbols = symbols.iter().map(|s| s.as_ref()).collect::<Vec<_>>();
```
isuffix:
```suggestion
        let symbols = ["a", "b", "c", "d", "e", "f", "g"];
```
This is starting to get more complex than it needs to be. There's no need to go all the way from a to z, so just using a static array should be fine.

Also, you can replace the two `symbols[i].to_string()` calls below with `symbols[i]` now that they're `&str`.

MDLC01: It's better indeed, thanks

isuffix: Aww, thanks for the coauthor credit :smiling_face_with_three_hearts:

isuffix (approved):

T0mstone (approved):

MDLC01:
Thank you both. I'll merge this this PR without waiting so that we can hopefully merge https://github.com/typst/codex/pull/170 before the next release.

For the record, this is not a breaking change for Typst at all.

PR #172, opened 2026-06-01
Update changelog for 0.3.0

MDLC01:
This reorganizes the existing changelog entries in a way that is hopefully easier to read quickly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Trivially updated to Unicode 17.0.0
 - Added numeral systems API under the `numeral-systems` feature flag
-  - Compared to Typst 0.14.2, added Tibetan numerals.
+  - Compared to Typst 0.14.2, added Tibetan numerals and Arabic letters in abjad order.
 
 ### New in `sym`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,41 +24,41 @@
   - `inter.serif`: ∩︀
   - `inter.sq.serif`: ⊓︀
   - `gt.double.nested`: ⪢
-  - `lt.double.nested`: ⪡
   - `gt.closed`: ⊳
   - `gt.closed.not`: ⋫
   - `gt.closed.eq`: ⊵
   - `gt.closed.eq.not`: ⋭
+  - `gt.arc` ⪧
+  - `gt.arc.eq` ⪩
+  - `gt.quest`: ⩼
+  - `lt.double.nested`: ⪡
   - `lt.closed`: ⊲
   - `lt.closed.not`: ⋪
   - `lt.closed.eq`: ⊴
   - `lt.closed.eq.not`: ⋬
-  - `supset.closed`: ⫐
-  - `supset.closed.eq`: ⫒
-  - `subset.closed`: ⫏
-  - `subset.closed.eq`: ⫑
-  - `gt.arc` ⪧
-  - `gt.arc.eq` ⪩
   - `lt.arc`: ⪦
   - `lt.arc.eq`: ⪨
+  - `lt.quest`: ⩻
+  - `supset.closed`: ⫐
+  - `supset.closed.eq`: ⫒
   - `supset.equiv`: ⫆
-  - `subset.equiv`: ⫅
   - `supset.nequiv`: ⫌
-  - `subset.nequiv`: ⫋
   - `supset.tilde`: ⫈
-  - `subset.tilde`: ⫇
   - `supset.approx`: ⫊
-  - `subset.approx`: ⫉
   - `supset.eq.dot`: ⫄
-  - `subset.eq.dot`: ⫃
   - `supset.plus`: ⫀
-  - `subset.plus`: ⪿
   - `supset.times`: ⫂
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -22,46 +25,41 @@
...
+  - `gt.closed.eq.not`: ⋭
+  - `gt.closed.not`: ⋫
+  - `lt.closed`: ⊲
+  - `lt.closed.eq`: ⊴
+  - `lt.closed.eq.not`: ⋬
+  - `lt.closed.not`: ⋪
+  - `subset.closed`: ⫏
+  - `supset.closed`: ⫐
+  - `supset.closed.eq`: ⫒
+  - `subset.closed.eq`: ⫑
```
T0mstone: I don't like the order of these four. For consistency with the lines above them, I'd suggest:
```md
  - `subset.closed`: ⫏
  - `subset.closed.eq`: ⫑
  - `supset.closed`: ⫐
  - `supset.closed.eq`: ⫒
```

CHANGELOG.md
```diff
@@ -22,46 +25,41 @@
...
   - `gt.double.nested`: ⪢
   - `lt.double.nested`: ⪡
+  - `gt.closed`: ⊳
+  - `gt.closed.eq`: ⊵
+  - `gt.closed.eq.not`: ⋭
+  - `gt.closed.not`: ⋫
+  - `lt.closed`: ⊲
+  - `lt.closed.eq`: ⊴
+  - `lt.closed.eq.not`: ⋬
+  - `lt.closed.not`: ⋪
```
T0mstone: Perhaps the simple `.not` should come before `.eq` and `.eq.not` in both cases.
Otherwise, it should at least come before `.eq.not`.

MDLC01:
@T0mstone I reordered what you suggested, and more. Tell me what you think.

CHANGELOG.md
```diff
@@ -22,46 +25,41 @@
...
-  - `supset.closed.eq`: ⫒
+  - `subset.approx`: ⫉
   - `supset.eq.dot`: ⫄
-  - `supset.equiv`: ⫆
-  - `supset.nequiv`: ⫌
+  - `subset.eq.dot`: ⫃
   - `supset.plus`: ⫀
-  - `supset.tilde`: ⫈
+  - `subset.plus`: ⪿
   - `supset.times`: ⫂
```
T0mstone: I think these should be grouped by `supset`/`subset` first. All other changelog entries also do it that way.

MDLC01: I don't think this is inconsistent: we first list all new `.closed` variants, then all new `.equiv` variant, then all new `.nequiv` variants... And every time, we list `supset` before `subset`.

T0mstone: But compare this to the `arrow` variants from 0.2.0: There, all `arrow.r` variants come before all `arrow.l` variants and so on. (Actually, with the exception of `arrow.l.open`, that seems to have been an oversight)
Similarly, the `plus.o` variants are all grouped together, and the `times.o` ones as well; We don't have `plus.o`, `times.o`, `plus.o.big`, `times.o.big`, `plus.o.l`, `times.o.l`, ...
In the same vein, I would have expected the `supset` and `subset` variants to be each grouped together, with the apparent exception that the closed symbols are listed separately because they form a more cohesive group like that (tho personally I would actually prefer them to just be grouped according to the base symbol, but I can accept this exception).

MDLC01: I grouped the entries by base symbol

MDLC01:
Arguably, `pee` should be listed under "Miscellaneous technical" and `mapsfrom.*` under "Arrows". What do you think?

T0mstone:
Not sure about `pee`: It doesn't really fit my conception of "technical", but maybe I'm misunderstanding what that means.
Similarly, I'd think it makes more sense for the tacks to be in "Mathematical symbols".
(Also, do we want to capitalize "Symbols" and "Technical"? The latter is capitalized in the 0.2.0 section)

For `mapsfrom.*`, I think either way has its benefits. They are arrows and literally just aliases for already existing `arrow` variants, so putting them under "Arrows" would make sense. On the other hand, I think these carry a bit more of an intent of use in formulas/equations, rather than diagrams, so having them under "Mathematical symbols" also makes sense.

MDLC01:
> Not sure about `pee`: It doesn't really fit my conception of "technical", but maybe I'm misunderstanding what that means.

Let's keep it there then, as it's mostly used for the Weierstrass elliptic function, which is a mathematical concept as far as I can tell.

> Similarly, I'd think it makes more sense for the tacks to be in "Mathematical symbols".

I would disagree with this as they are mostly used in formal logic, which I don't think is strictly a mathematical field.

> Also, do we want to capitalize "Symbols" and "Technical"?

We should only use title case for actual headings imo.

> On the other hand, I think these carry a bit more of an intent of use in formulas/equations, rather than diagrams, so having them under "Mathematical symbols" also makes sense.

I agree with this.

T0mstone (approved):

laurmaedje:
Thanks for keeping the changelog up to date. It really makes things much easier for me!

PR #174, opened 2026-06-16
Add missing tacks (part II)

MDLC01:
This is a follow-up to #151.

Some of the names we use in this PR were made available for this purpose by the deprecations introduced in #151.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,35 @@
 
 ## Unreleased
 
-### Removals in `sym` **(Breaking change)**
-These previously deprecated items were removed:
-- `gt.tri.*`
-- `lt.tri.*`
-- `join`, `join.*`
+## New in `sym`
+
+- Miscellaneous technical
+  - `tack.r.double` ⊩
+  - `tack.r.double.not` ⊮
+  - `tack.r.triple` ⊪
+  - `tack.rr.short` ⊧
+  - `tack.rr.double` ⊫
+  - `tack.rr.double.not` ⊯
+  - `tack.l.double` ⫣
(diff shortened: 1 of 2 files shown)


isuffix (approved):
Looks great

sijow (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+## New in `sym`
+
+- Miscellaneous technical
+  - `tack.r.double` ⊩
```
T0mstone, line 8: I think it makes sense as is, but perhaps these should use a "Changed codepoint" section like we've done in the past, instead of being removed and added at the same time?

MDLC01: What I don't like with a "Changed codepoints" section is that it buries what are effectively additions under a different part.

What about listing those in "New in `sym`" like now, but adding a "Changed codepoints" section as well?

T0mstone: That's not a bad idea. I'd actually want to *move* them there from the "Removals" then tho, i.e. have them in "New in `sym`" and in "Changed codepoints (breaking change)", but not in "Removals in `sym` (breaking change)"

MDLC01: I applied the change

MDLC01:
This is missing a review to be merged because of the changelog changes

sijow (approved):

isuffix (approved):

MDLC01:
This needs one last review 😭

sijow (approved):

MDLC01:
Thank you!

PR #178, opened 2026-07-06
Add ordinal ª and º indicators

sdalmonte:
This pull request adds the [ordinal indicators](https://en.wikipedia.org/wiki/Ordinal_indicator) ª and º.

| character | Unicode value | symbol |
| - | - | - |
| ª | U+00AA FEMININE ORDINAL INDICATOR | `ordinal.a` |
| º | U+00BA MASCULINE ORDINAL INDICATOR | `ordinal.o` |

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -732,9 +732,10 @@ suit
   .heart.stroked ♡
   .spade.filled ♠\vs{text}
   .spade.stroked ♤
-ordinal
-  .a ª
-  .o º
+ordinal {
+  a ª
+  o º
+}
 
 // Music.
 note
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -732,6 +732,9 @@ suit
   .heart.stroked ♡
   .spade.filled ♠\vs{text}
   .spade.stroked ♤
+ordinal
+  .a ª
+  .o º
 
 // Music.
 note
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -733,8 +733,8 @@ suit
   .spade.filled ♠\vs{text}
   .spade.stroked ♤
 ordinal {
-  a ª
-  o º
+  feminine ª
+  masculine º
 }
 
 chess {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### New in `sym`
+
+- Ordinal indicators
+  - `ordinal.a`: ª
+  - `ordinal.o`: º
+
 ### Removals in `sym` **(Breaking change)**
 These previously deprecated items were removed:
 - `gt.tri.*`


src/modules/sym.txt
```diff
@@ -732,6 +732,9 @@ suit
   .heart.stroked ♡
   .spade.filled ♠\vs{text}
   .spade.stroked ♤
+ordinal
+  .a ª
+  .o º
```
```suggestion
ordinal {
  a ª
  o º
}
```

I think we should use a module instead, so that `ordinal` alone does not fall back to an arbitrary value.

MDLC01 (commented):
For reference, those characters are documented in the Unicode Core Specification.[^1] Notably, they do not seem to be included for compatibility reasons, which motivates their addition to Codex.

[^1]: https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-7/#G6207

MDLC01:
You will also need to update the changelog in CHANGELOG.md

MDLC01 (approved):

mkorje (approved):

MDLC01:
We are trying a new way to keep track of changes by maintaining the changelog in a separate shared document (https://typst.app/project/r2ChBGT9XWBoEhAKGM6icb). Feel free to drop the update to the changelog from this PR to prevent any further conflict.

Enivex:
Would it make more sense to call them `ordinal.f` and `ordinal.m`? For reference, the names in HTML are `&ordf;` and `&ordm;`, they are `\textordfeminine` and `\textordmasculine` in latex, and the unicode names also use feminine/masculine. I don't see a particularly good reason for typst to go against the grain here.

MDLC01:
One of the advantages of using `.a` and `.o` is that it is more future-proof in case one day Unicode adds ordinal indicators for other languages that use different letters. For this reason, I am in favor of keeping `.a` and `.o`.

If we go for a more semantic name, I would be in favor of longer modifiers, such as `.fem` and `.masc`, if not `.feminine` and `.masculine`. For comparison, we already use similarly long modifiers, e.g., `gender.female`.

Enivex:
> One of the advantages of using `.a` and `.o` is that it is more future-proof in case one day Unicode adds ordinal indicators for other languages that use different letters. For this reason, I am in favor of keeping `.a` and `.o`.

I'm not sure I buy that argument. They could equally well add ordinal symbols that visually look the same, but carry different semantic information.

> If we go for a more semantic name, I would be in favor of longer modifiers, such as `.fem` and `.masc`, if not `.feminine` and `.masculine`. For comparison, we already use similarly long modifiers, e.g., `gender.female`.

I don't think `fem` and `masc` would be very consistent with other naming, so in that case I'd prefer the full name

MDLC01:
Then maybe we can have a consensus on `.feminine` and `.masculine`?

sdalmonte:
With all that’s been shared here, `.feminine` and `.masculine` make more sense to me as well, for the following reasons:

- There’s a consensus between Unicode, HTML entities and LaTeX to refer to the ordinal characters semantically, not visually, which rules out `.a` and `.o`.
- Codex tends to use single-character suffixes for visual, not semantic purposes, which rules out `.f` and `.m`.
- Codex also tends not to use abbreviations, which rules out `.fem` and `.masc`.

I’ll update my pull request.

MDLC01 (approved):

sijow (approved):

PR #179, opened 2026-07-06
Add Canadian French trademark symbols

### LINKED ISSUE #177: Add Canadian French trademark symbols
In Canada, the ™ and ® symbols have French-language equivalents, 🅪 (`U+1F16A RAISED MC SIGN`) and 🅫 (`U+1F16B RAISED MD SIGN`) respectively. I propose adding them as variants of `trademark` and `trademark.registered`, with a common suffix to indicate that their use is specific to French-language materials in Canada. 

There are currently no symbols in `sym.txt` that are both region- and language-specific. Following the lead of variants of the `peso` and `rupee` currency symbols, the suffix could be `.canadianfrench` (or `.canadian.french`), but that’s quite long. For brevity I would suggest `.frca`, based on the ISO language code.

sdalmonte:
Closes #177.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 ## Unreleased
 
-### New in `sym`
-
-- Canadian French trademark symbols
-  - `trademark.mc`: 🅪
-  - `trademark.md`: 🅫
-
 ### Removals in `sym` **(Breaking change)**
 These previously deprecated items were removed:
 - `gt.tri.*`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### New in `sym`
+
+- Canadian French trademark symbols
+  - `trademark.mc`: 🅪
+  - `trademark.md`: 🅫
+
 ### Removals in `sym` **(Breaking change)**
 These previously deprecated items were removed:
 - `gt.tri.*`
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -722,6 +722,8 @@ copyleft 🄯
 trademark ™\vs{text}
   .registered ®\vs{text}
   .service ℠
+  .mc 🅪
+  .md 🅫
 maltese ✠
 suit
   .club.filled ♣\vs{text}


MDLC01:
I would like to see other name ideas. Right now, I don't really like the inconsistency between the Canadian French trademark symbols, which use their abbreviations, and the other trademark symbols, which use semantic names.

I would like something along the lines of `trademark.commerce` and `trademark.deposee`, although I don't really like that this uses non-English words, especially for `trademark.deposee` which just doesn't read right in any language.

sdalmonte:
I feel like any option requires some tradeoffs. Here's a few ideas:

| symbols | pros | cons |
| - | - | - |
| `trademark.mc`<br>`trademark.md` | concise, describes the appearance of the symbol | not semantic; doesn't make it clear that this is a region-specific symbol |
| `trademark.commerce`<br>`trademark.deposee` | semantic (describes a French-only usage in French) | we don't currently use non-English terms in symbols; "deposee" is technically incorrect (should be "déposée", but symbols shouldn't contain accents); doesn't make it clear that this is a region-specific symbol |
| `trademark.marquedecommerce`<br>`trademark.marquedeposee` | semantic (describes a French-only usage in French) | we don't currently use non-English terms in symbols; "deposee" is technically incorrect (should be "déposée", but symbols shouldn't contain accents); doesn't make it clear that this is a region-specific symbol; extremely long |
| `trademark.canadianfrench`<br>`trademark.registered.canadianfrench` | matches the semantics of the other symbols under `trademark` and localized currency symbols (e.g. `peso.philippine`, `rupee.tamil`) | extremely long |
| `trademark.canadian.french`<br>`trademark.registered.canadian.french` | matches the semantics of the other symbols under `trademark` and localized currency symbols (e.g. `peso.philippine`, `rupee.tamil`) | extremely long, potentially introduces ambiguity since there's no `trademark.french` / `trademark.canadian` symbol |
| `trademark.frca`<br>`trademark.registered.frca` | semantic (using ISO locale code) | use of abbreviations is inconsistent with other locale/language-specific symbols |
| `trademark.fr.ca`<br>`trademark.registered.fr.ca` | semantic (using ISO locale code) | use of abbreviations is inconsistent with other locale/language-specific symbols; potentially introduces ambiguity since there's no `trademark.fr` / `trademark.ca` symbol |
| `marquedecommerce`<br>`marquedeposee` | spells out the abbreviation of the symbol | quite long; since it's not within `trademark`, we lose the semantics of `trademark`; we don't currently use non-English terms in symbols; "deposee" is technically incorrect (should be "déposée", but symbols shouldn't contain accents); doesn't make it clear that this is a region-specific symbol |

Of these, I prefer `trademark.frca`/`trademark.registered.frca` or `trademark.commerce`/`trademark.deposee`. I'm open to other suggestions.

MDLC01:
I don't really like options that involve specifying a locale. I am not sure why, but I don't like it. I also think we should have those symbols as variants of `trademark` and prevent extremely long names. So out of all the options you specified, I would be in favor of the `trademark.commerce`/`trademark.deposee` combo or the `trademark.mc`/`trademark.md` combo.

I don't think there is a technical reason why we couldn't use `trademark.déposée`,[^1] but I agree we probably don't want to introduce accented letters in symbol names.

[^1]: We would need to check how canonically equivalent identifiers are handled by Typst because there are multiple ways to write "déposée" using Unicode, but that is not a technical limitation.

sdalmonte:
In that case I still prefer `trademark.mc` and `trademark.md`. They’re less semantic, but they’re concise and match the appearance of the 🅪 and 🅫 symbols themselves.

That said, I would welcome other opinions, especially from Canadian francophones.

Enivex:
Of all options, I think `trademark.commerce` and `trademark.deposee` makes the most sense. That being said, I do not speak French.

sijow:
Noncanadian francophone here... Reading something that sounds just like "marque déposée" but with the "wrong" spelling for the first word is weird but I think it's still my preferred option. I would prefer `trademark.mc` and `trademark.md` if that was consistent with existing symbols, but it's not (or `trademark.registered` should be `trademark.r`)

vmartel08:
As a French Canadian, I think `trademark.md` and `trademark.mc` are the most concise, clear, and self-explanatory names for someone who wants to insert those symbols.

Alternatives such as `trademark.déposée` and `trademark.commerce` works for me as well. I would steer clear of `trademark.deposee`.

I can hardly see any benefit of including the locale such as `frca`.

MDLC01:
There doesn't seem to be a general consensus, but the two Canadian francophones have expressed opinions in favor of `trademark.md` and `trademark.mc`, so I think we can go with this option.

You can remove all changes to the changelog to prevent any further conflict.

sdalmonte:
Done! Thanks @MDLC01.

MDLC01 (approved):

sijow (approved):

Issue #108, opened 2025-07-04
Find consistent way to refer to vertical mirroring


Enivex:
We seem to only have two current conventions:

`.rev` - Horizontal mirroring
`.inv` - Rotate by 180 degrees

These can also be combined to yield

`.rev.inv`- Vertical mirroring

There are several drawbacks with the current situation:
- The word "invert" can also imply vertical mirroring. In fact, this is the sense in which it is used in unicode naming (though inconsistently).
- If there is only a vertically mirrored variant, then using `.rev.inv` isn't exactly the most natural modifier. This is the case for obtuse angles, where ⦦ (currently `angle.oblique`) and ⦧ (not currently named) are the only variants.

I see several possibilities
- Introduce `.turned` or `rot` (though this has also been proposed for 90 degrees) for rotation by 180 degrees, and use `.inv` for vertical mirroring
- Keep `.inv` for rotation by 180 degrees, and use `.flip` for vertical mirroring
- Something like `.rev.h` and `.rev.v` (though I can imagine this leading to conflicts if we don't have #94 #95 )

MDLC01:
I agree that the `.rev.inv` situation is unfortunate.

I also agree that `.rev` and `.inv` are not necessarily the clearest names, but I feel like they work anyway for two reasons: they are short, and they still convey the general idea. When reading Typst source code, it's probably not that bad if the modifiers don't make it perfectly clear exactly the way a symbol is rotated/mirrored. When writing Typst code this is more of an issue as you need to remember which modifier gives you the symbol you want, but you only need to figure this once per symbol (then I feel like most people can remember the names of the symbols they use frequently).

Regarding your specific propositions:
- `.turned` is too long imo. Additionally, `.flip` would have to be `.flipped` for consistency (https://github.com/typst/codex/issues/43#issuecomment-3037212874).
- To be clear, that would just mean changing `.rev.inv` to `.flip`? If so, I like that. Although there is again the issue of "flip" vs. "flipped".
- I don't like that because `.h` and `.v` are widely used with a meaning that applies to the symbol as a whole.

Regarding the past participle issue ("flip" vs. "flipped"), maybe we can accept "flip" as an abbreviation of "flipped", the same way "rot" is an abbreviation of "rotated".

Enivex:
For symbols that are horizontally symmetric, should `.inv` or `.flip` be the default way to refer to the upside down variant?

This currently affects three symbols ⸸ ¡ ℧, but more in the future.

[...]

PR #180, opened 2026-07-12
Add `not.inv` and `not.rev`

MDLC01:
`not.inv` is consistent with `quest.inv` (¿). `not.rev` is consistent with `in.rev` (∋).

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -536,6 +536,8 @@ exists ∃
 top ⊤
 bot ⊥
 not ¬
+  .inv ⌙
+  .rev ⌐
 and ∧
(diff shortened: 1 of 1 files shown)


T0mstone (approved):

sijow:
This relates to https://github.com/typst/codex/issues/108 . If a short term solution is desired I'm OK but for long term I would favor renaming our current `inv` to `turned`.

MDLC01:
The symbols added in this PR are not widely used AFAIK, but they fit in the current naming schema. Progress on #108 can happen independently.

sijow (approved):

PR #181, opened 2026-07-22
Add more stacked tilde variants

MDLC01:
This adds a few symbols composed of stacked tildes and horizontal lines. I explain the reasoning for each change below.

- `tilde.double` makes sense in my opinion as a non-semantic variant of `approx`. I have chosen to only include the `.not` variant but I am open to including `.eq`, `.equiv`, and `.hat` as well. I am also open to excluding `tilde.double` from this PR if we are unable to reach a consensus.

- The rename from `minus.tilde` to `eq.tilde` is consistent with the existing `tilde.eq`, as well as the Unicode character being a Relation, not a Binary, even though its name uses "MINUS" instead of "EQUAL". Same for `tilde.eq.tilde`.

- `equiv.tilde` is consistent with the existing `tilde.equiv`.

- `approx.equiv` is consistent with other uses of `.equiv`, such as in `tilde.equiv`.

### DIFF
--- a/src/modules/sym.txt
+++ b/src/modules/sym.txt
@@ -196,9 +196,12 @@ tilde
   .op ∼
   .basic ~
   .dot ⩪
+  .double ≈
+  .double.not ≉
   .eq ≃
   .eq.not ≄
   .eq.rev ⋍
+  .eq.tilde ⩬
   .equiv ≅
   .equiv.not ≇
   .nequiv ≆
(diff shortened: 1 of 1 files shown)


MDLC01:
This PR is missing three approvals to be merged. If you disagree with a specific change, feel free to express your concern here or on Discord.

sijow (approved):

Issue #11, opened 2023-11-24
Filled circles are out of order


Enivex:
### Description

According to https://www.unicode.org/charts/PDF/U1F780.pdf are ordered by size as follows:
22C5 < 2219 <1F784 <  2022 < 2981 < 26AB < 25CF < 2B24

But in typst U+2219 is circle.filled small and U+2981 is circle.filled.tiny. I suspect this is because the font used in the symbol list, but better fonts that actually follow unicode properly (Stix Two Math for instance) shows `$ circle.filled.tiny circle.filled.small$` as 
![image](https://github.com/typst/typst/assets/2191754/506d02d9-3192-460e-8ddc-8bebe2ab8ba8)

(This could also apply to other families of symbols, but this is the one I noticed)

### Typst version

- [X] I am using the latest version of Typst

MDLC01:
For reference, a list of most geometric shapes, which also seems to be in order, is available in [Table 22-7 of the Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G57602).

Enivex:
> For reference, a list of most geometric shapes, which also seems to be in order, is available in [Table 22-7 of the Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G57602).

That page also explicitly says 

 "Several other blocks, such as General Punctuation, Mathematical Operators, Block Elements, Miscellaneous Symbols, and Geometric Shapes Extended, contain a few other characters which are members of the **size-graded sets of such symbols**."

MDLC01:
Additional information regarding the use of geometric shapes in mathematics, including available sizes and size order, is available in section 2.11 of [UTR #25](https://www.unicode.org/reports/tr25/), the (non-authoritative) Technical Report on Unicode Support for Mathematics.

PR #182, opened 2026-08-12
Add `space-discarding` feature

isuffix:
This adds the `space-discarding` feature as described at https://github.com/typst/typst/pull/7350#issuecomment-5122759812, although I have changed from **"whether a writing system uses spaces between words"** to **"whether a writing system uses spaces at all."**

I will leave this PR description short as the code itself contains a plenty of discussion of rationale and implementation considerations, along with my research into the usage of space characters in various writing systems.

There is a lot of writing here, so I would really appreciate help with checking for typos and inconsistencies, as it has become hard for me to consider everything with fresh eyes. I am very amenable to suggestions :)

I will also restate that I only speak English and while I have tried to do good research, I am not infallible. I would appreciate any input from native speakers of Chinese or Japanese or any of the other writing systems discussed in the PR.

I would also like to thank @r12a for his wonderfully detailed [orthography descriptions](https://r12a.github.io/scripts/index.html#scriptnotes) and [script comparison table](https://r12a.github.io/scripts/script-features/index.html), without which this PR would not be nearly as complete or authoritative. If you're reading this, I would love any feedback you could provide.


r12a:
Tibetan comes to mind as an example of an orthography that doesn't use (ASCII) spaces but does have delimiters (syllable-based) which could break a line in the source code, but which should not incur an extra space when stitching things together (see [Tibetan Orthography Notes](https://r12a.github.io/scripts/tibt/bo.html#linebreak)). hth

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+use icu_properties::props::{EastAsianWidth as EAW, Emoji, Script};
+use icu_properties::{CodePointMapDataBorrowed, CodePointSetDataBorrowed};
+
+/// Whether to discard an inferred space between two strings. This is `true`
+/// when the inner character of _either_ string is unambiguously from a writing
+/// system which does not use spaces. Otherwise the inferred space should be
+/// kept.
+///
+/// Currently this check includes characters which we determine to be from the
+/// Chinese, Japanese, or Yi writing systems plus ideographic punctuation. Note
```
YDX-2147483647, line 56: The term _ideographic punctuation_ needs clarification.

- There's a Unicode block called [_Ideographic Symbols and Punctuation_](https://www.unicode.org/charts/PDF/U16FE0.pdf), but obviously that isn't what you mean.

- `‼` is listed in both [Punctuation marks in Chinese in CLReq](https://www.w3.org/TR/clreq/#tables_of_chinese_punctuation_marks) and [Character Classes in JLReq](https://www.w3.org/TR/jlreq/#character_classes), but it isn't considered as a _ideographic punctuation_.

It looks like that you use this term colloquially. I suggest putting the relevant test cases in a separate function and linking to it. (similar to `test_spacing_emoji_presentation`)

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+        check_spacing('＿', No);
+        check_spacing('﹏', No);
+        // Not these punctuation though:
+        check_spacing('#', YesOrAmbiguous);
+        check_spacing('*', YesOrAmbiguous);
+        check_spacing('/', YesOrAmbiguous);
+        check_spacing('-', YesOrAmbiguous); // hyphen
+        check_spacing('–', YesOrAmbiguous); // en-dash
+        check_spacing('—', YesOrAmbiguous); // em-dash
+        check_spacing('⸺', YesOrAmbiguous); // two em-dash
```
YDX-2147483647:
```suggestion
        check_spacing('–', YesOrAmbiguous); // en dash
        check_spacing('—', YesOrAmbiguous); // em dash
        check_spacing('⸺', YesOrAmbiguous); // two-em dash
```

It should be _two-em dash_, not _two em-dash_.

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+/// wordspace characters instead of using the dedicated codepoint.
+///
+/// The wordspace character generally attaches to the end of words when spacing
+/// is added (such as for justification), and linebreaks should never separate
+/// the wordspace from its preceding character. When using the wordspace
+/// character, justified text can insert whitespace either on both sides of the
+/// wordspace, centering it between words, or immediately after the wordspace,
+/// leaving it connected to the previous word.
+///
+/// We should avoid intoducing space characters that the author didn't intend
```
YDX-2147483647:
```suggestion
/// We should avoid introducing space characters that the author didn't intend
```

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+    YesOrAmbiguous,
+}
+
+/// Determine the [`WritingSystemSpacing`] for a character.
+///
+/// It was very useful to consult the `UnicodeSet` utility when defining this
+/// function: <https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp>.
+///
+/// We check the East Asian Width property being `Fullwidth`, `Halfwidth`, or
+/// `Wide` to determine common ideographic punctuation characters, unfortunately
```
YDX-2147483647:
```suggestion
/// `Wide` to determine common ideographic punctuation characters. Unfortunately,
```

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+        // Like Firefox, we treat the Korean won sign as ambiguous, although not
+        // the fullwidth variant. This matches the behavior of the Yen/Yuan sign
+        // and its fullwidth variant, although Chinese and Japanese usually use
+        // a different character instead of `¥`. Note that Latin text uses `¥`
+        // for currency from both Chinese and Japanese, which often requires
+        // disambiguation. What fun!
+        check_spacing('₩', YesOrAmbiguous);
+        check_spacing('￦', No);
+        check_spacing('¥', YesOrAmbiguous);
+        check_spacing('￥', No);
```
YDX-2147483647: Just a comment for your information: I get the narrow ¥ when I type in Chinese mode on my mobile phone, but I get the fullwidth ￥ when I press Shift+4 ($) in Chinese mode on my desktop computer.

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
+//! Whether to keep or discard spaces that are inferred due to newlines in
```
YDX-2147483647: I think it's necessary to explain the meaning of _discarding spaces_ here, as the word _space_ is also used to refer to the _lack of ink_ below in `WritingSystemSpacing`.

My suggestion:
- Refer to the lack of ink as _spacing_.
- Refer to U+0020 SPACE and `\n` as _whitespace characters_.

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+        assert!(discard_space_between("", "漢"));
+        // Keep (note the `!`)
+        assert!(!discard_space_between("", ""));
+        assert!(!discard_space_between("A", ""));
+        assert!(!discard_space_between("", "A"));
+
+        // Spaces themselves don't get special treatment
+        assert!(discard_space_between("漢", " "));
+        assert!(discard_space_between(" ", "漢"));
+        assert!(!discard_space_between(" ", " "));
```
YDX-2147483647: Is there any use case for empty strings and spaces?

If there is, we should mention it in the public doc comments.
Otherwise, perhaps we should make `discard_space_between(a, b)` always equivalent to `discard_space_between(a.trim(), b.trim())`. That might be useful when dealing with the following markdown.

```markdown
- 比如
  这样
- The second
  point
```

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+        check_spacing('〖', No);
+        check_spacing('〗', No);
+        check_spacing('〔', No);
+        check_spacing('〕', No);
+        check_spacing('［', No);
+        check_spacing('］', No);
+        check_spacing('｛', No);
+        check_spacing('｝', No);
+        check_spacing('＿', No);
+        check_spacing('﹏', No);
```
YDX-2147483647, line 346: Is it possible to replace these `check_spacing(.., No)` lines with something like the following? This can prevent `writing_system_spacing` from discarding spaces around punctuations accidentally.

```rust
assert_eq!(
  (0..=0x10FFFF)
    .filter_map(char::from_u32)
    .filter(|c| c.is_puncutation() && matches!(writing_system_spacing(c), No)),
  &[..],
);
```

YDX-2147483647 (changes requested):
I've gone through the source code (except the emoji part). I believe the current algorithm is simple and robust enough. I only have some suggestions regarding doc comments and tests. See my individual comments.

## More materials supporting the current algorithm

### East Asian Width

[UAX #14: Unicode Line Breaking Algorithm](https://www.unicode.org/reports/tr14/#Algorithm) also uses East Asian Width to filter out East Asian characters.

> The symbol `$EastAsian` stands for the set `[\p{ea=F}\p{ea=W}\p{ea=H}]` of characters with Fullwidth, Wide, or Halfwidth East Asian Width.

### Pandoc 

Pandoc's [`east_asian_line_breaks`](https://pandoc.org/MANUAL.html#extension-east_asian_line_breaks) extension [uses `charWidth`](https://github.com/jgm/pandoc/blob/1552249b903651d181e1cb2450b5b1860d5eb03d/src/Text/Pandoc/Transforms.hs#L137-L149) to determine if a soft break (e.g., single newline in markdown and typst) should be removed. However, we've argued in https://github.com/typst/typst/pull/7350#discussion_r2576402285 that the [rules for determining widths](https://docs.rs/unicode-width) are too complicated and it's better to use East Asian Width directly.

### LaTeX

LaTeX cannot be taken as a reference, because the implementations are limited by the technology. Specifically, the whitespace in `字\n“` should be discarded, but luatexja keeps it. And the whitespace in `”\nA` should be kept as a word space, but xeCJK discards it. See https://github.com/typst/typst/issues/792#issuecomment-2303677367 for the tests.

### Typst cjk-unbreak

As for typst packages, [cjk-unbreak uses the following regex](https://github.com/KZNS/cjk-unbreak/blob/2ea9b0ce3654ab537116499f63aa4077165192bc/remove-cjk-break-space.typ#L4-L20) to determine if a character is CJ (Chinese + Japanese) and discards the space iff either side matches the regex.

```
[\p{Han}，。；：！？‘’“”（）「」【】…—\p{Hiragana}\p{Katakana}]
```

This package is designed only for CJ, so the regex includes a few characters that are considered `YesOrAmbiguous` in this PR.
The algorithm in this PR is designed for all writing systems, so the difference to cjk-unbreak is acceptable.

### Typst cjk-spacer

A newer typst package, cjk-spacer, [uses a more complex algorithm](https://github.com/ryuryu-ymj/cjk-spacer/blob/327e21f568ed902b341cfa506904c0b7d8fc5b36/src/lib.typ#L1-L91). If I understand correctly, then its algorithm is equivalent to the following.

```typst
#let default-cjk-regex = regex(
  "["
    + "\p{scx:Hira}\p{scx:Kana}\p{scx:Han}\p{scx:Hang}\p{scx:Bopo}"
    + "\u3000-\u303F" // CJK Symbols and Punctuation
    + "\u3190-\u319F" // Kanbun
    + "\u31C0-\u31EF" // CJK Strokes
    + "\u3200-\u32FF" // Enclosed CJK Letters and Months
    + "\u3300-\u33FF" // CJK Compatibility
    + "\uFE10-\uFE1F" // Vertical Forms
    + "\uFE30-\uFE4F" // CJK Compatibility Forms
    + "\uFE50-\uFE6F" // Small Form Variants
    + "\uFF00-\uFFEF" // Halfwidth and Fullwidth Forms
    + "]",
)
#let default-western-open-punc-regex = regex(
  "[\p{Pi}\p{Ps}--["
    + "\u3000-\u303F" // CJK Symbols and Punctuation
    + "\uFE10-\uFE1F" // Vertical Forms
    + "\uFE30-\uFE4F" // CJK Compatibility Forms
    + "\uFE50-\uFE6F" // Small Form Variants
    + "\uFF00-\uFFEF" // Halfwidth and Fullwidth Forms
    + "]]",
)
#let default-western-close-punc-regex = regex(
  "[\p{Pf}\p{Pe}\p{Term}--["
    + "\u3000-\u303F" // CJK Symbols and Punctuation
    + "\uFE10-\uFE1F" // Vertical Forms
    + "\uFE30-\uFE4F" // CJK Compatibility Forms
    + "\uFE50-\uFE6F" // Small Form Variants
    + "\uFF00-\uFFEF" // Halfwidth and Fullwidth Forms
    + "]]",
)

#let discard_space_between(before, after) = {
  if after.matches(western-open-punc-regex).len() == 0 and after.starts-with(cjk-regex) {
    true
  } else if before.matches(western-close-punc-regex).len() == 0 and before.ends-with(cjk-regex) {
    true
  } else {
    false
  }
}
```

The cjk-spacer algorithm does not merely consider the characters _immediately adjacent_ to the space, but rather the _text segments_ around the space. This approach is appropriate when typesetting a document, but it's too surprising at the syntax level.

Also, cjk-spacer treats K the same as CJ. According to previous feedbacks in https://github.com/typst/typst/pull/7350, discarding spaces is not preferable for Korean texts.

And I haven't check if the Unicode blocks enumerated by cjk-spacer are equivalent to this PR, but I think it's worth checking before merging this PR.

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+/// kept.
+///
+/// Currently this check includes characters which we determine to be from the
+/// Chinese, Japanese, or Yi writing systems plus ideographic punctuation. Note
+/// that Korean does use spaces between words and predominantly uses Latin
+/// punctuation instead of ideographic punctuation.
+///
+/// We currently do not discard inferred space characters from writing systems
+/// which use spaces only between phrases or sentences. This is because there
+/// are not clear distinctions for where phrases or sentences may or may not
```
laurmaedje:
```suggestion
/// are no clear distinctions for where phrases or sentences may or may not
```

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+///
+/// These systems do use empty space in writing, but it is empty due to the lack
+/// of ink in ideographic characters typeset on a grid (like the ideographic
+/// fullstop `。`), not due to explicit space characters in text.
+///
+/// Some ideographic punctuation characters primarily used by Chinese, Japanese,
+/// and Yi are also used in Korean writing (which uses spaces between words).
+/// But these are often used in vertical writing whereas modern Korean
+/// predominantly uses Latin punctuation when written horizontally. So this
+/// definition considers ideographic punctuation characters as from writing
```
laurmaedje: the phrasing with "as from" reads a bit confusingly to me

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+//! cites Chinese and Japanese for clarity, we endeavor to have ergonomic
+//! behavior for all writing systems, scripts, and languages.
+//!
+//! Much of the research on writing systems used for this feature was based on
+//! the wonderfully detailed [Orthography Descriptions][scriptnotes] by Richard
+//! Ishida. They proved invaluable here, and are certainly helping his mission
+//! of making the World Wide Web worldwide :)
+//!
+//! [scriptnotes]: https://r12a.github.io/scripts/index.html#scriptnotes
+use icu_properties::props::{EastAsianWidth as EAW, Emoji, Script};
```
laurmaedje:
```suggestion

use icu_properties::props::{EastAsianWidth as EAW, Emoji, Script};
```
I'm also not sure about the `as EAW` rename. It feels to me like it reduces clarity, if anything.

Cargo.toml
```diff
@@ -12,12 +12,14 @@ categories = ["encoding", "text-processing"]
...
 [features]
-default = ["numeral-systems", "styling"]
+default = ["numeral-systems", "styling", "space-discarding"]
 numeral-systems = ["dep:chinese-number"]
 styling = []
+space-discarding = ["dep:icu_properties"]
 _test-unicode-conformance = ["dep:ureq", "dep:rustls-webpki"]
 
 [dependencies]
+icu_properties = { version = "2.2.0", optional = true }
```
laurmaedje: please sort dependencies alphabetically

src/space_discarding.rs
```diff
@@ -0,0 +1,462 @@
...
+        // Discard
+        assert!(discard_space_between("漢", "漢"));
+        assert!(discard_space_between("1", "漢"));
+        assert!(discard_space_between("漢", "A"));
+        assert!(discard_space_between("漢。", "A"));
+        assert!(discard_space_between("A", "（漢"));
+        assert!(discard_space_between("A。", "B"));
+        assert!(discard_space_between("A", "（B"));
+
+        // Keep (note the `!`)
```
laurmaedje: The tests could be made clearer by adding a test-local `fn keep_space_between` and dropping the extra negations.

laurmaedje (commented):
Logic-wise, this seems reasonable and very well supported in terms of rationale. I'd trust in the research and the other involved people's judgement. I just have some minor nits.

src/space_discarding.rs
isuffix: I'll remove the rename. It was more helpful when I had all of the `EastAsianWidth` variants in one `matches!()` call.

src/space_discarding.rs
isuffix: Rewriting this sentence to: _"So this definition treats these shared characters as part of the writing systems which do not use spaces."_

src/space_discarding.rs
isuffix: We could, it's just a bit much to enumerate 792 codepoints (or 59 codepoint ranges).

You can explore the codepoints using the [UnicodeSet](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp) utility, [here's the set](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BCommon%7D+%26+%5B%5Cp%7BEA%3DH%7D%5Cp%7BEA%3DF%7D%5Cp%7BEA%3DW%7D%5D+-+%5Cp%7BEmoji%7D+-+%5B%E2%82%A9%5D&g=&i=Scx) of punctuation we're currently discarding around: `\p{Common} & [\p{EA=H}\p{EA=F}\p{EA=W}] - \p{Emoji} - [₩]`.
 (I got 59 ranges by using Ctrl+F for "items:")

src/space_discarding.rs
isuffix: Lexing already groups adjacent whitespace with newlines, so the markup `A    \n    B` will be treated exactly like `A\nB`. The exception is comments: `A //\nB`, but my implementation handles those by collapsing adjacent space elements into one before calling `discard_space_between`.

I would prefer to have the caller apply `trim` themselves so that this function doesn't do any unecessary work for Typst's case.

src/space_discarding.rs
isuffix: I'll split into two test functions and integrate the "Miscellaneous" section into the two punctuation groups.

Do you think we should consider special-casing `⁇` or `‼`?

src/space_discarding.rs
isuffix: I'll change all of the introductory sentences for doc comments to make sure they use _space characters_, but I don't I don't really want to read _whitespace characters_ everywhere it currently says _spaces_. I think as long as the introductions are consistent the usage should be clear.

isuffix:
I've updated comments and tests based on all of the feedback!

I think for this initial PR we will not include Tibetan. But one of the main points for moving this to Codex is so we can add GitHub issues and iterate at a separate pace than `typst/typst`. So we should add an issue for Tibetan support and discuss it more once this is merged.

For cjk-unbreak, it looks there are quite a few differences to this PR. The characters it doesn't include are all either `Yi` only or are `ScriptExtensions=Common` that have `EastAsianWidth=Wide`. For the other way around, this [UnicodeSet link] grouped by `EastAsianWidth` with `Script=Hangul` removed, shows 89 characters.

Of those, it seems the groups we should consider are:
1. The fullwidth latin letters: U+FF21-U+FF5A (Ａ - ｚ)
2. The circled numbers on black squares: U+3248-U+324F (㉈-㉏)
3. The [six CJ marks] with `Sc=Inherited` and `EA=W`: U+302A-U+302D and U+3099-U+309A
4. The [corner tone marks]: U+A700-U+A707 ([Wikipedia](https://en.wikipedia.org/wiki/Modifier_Tone_Letters))
5. [U+303F] IDEOGRAPHIC HALF FILL SPACE

I'll prepare an update to include these.

<details><summary>UnicodeSet syntax for this PR and cjk-unbreak</summary>
<p>

This PR currently:
```regex
[
  \p{Han} \p{Kana} \p{Hira} \p{Bopo} \p{Yi}
  [ \p{Common} & [\p{EA=H}\p{EA=F}\p{EA=W}] - \p{Emoji} - [₩] ]
]
```

cjk-unbreak default (excluding `\p{Scx=Hang}`):
```regex
[
  \p{Scx=Hira} \p{Scx=Kana} \p{Scx=Han} \p{Scx=Bopo}
  [ \u3000-\u303F \u3190-\u319F \u31C0-\u31EF \u3200-\u32FF \u3300-\u33FF \uFE10-\uFE1F \uFE30-\uFE4F \uFE50-\uFE6F \uFF00-\uFFEF ]
]
```

</p>
</details>

[corner tone marks]: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BNames_List_Subheader%3DCorner+tone+marks+for+Chinese%7D&g=&i=Sc%2CScx%2CEA

[U+303F]: https://util.unicode.org/UnicodeJsps/character.jsp?a=303F

[six CJ marks]: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B+%5Cp%7BNames_List_Subheader%3DCombining+tone+marks%7D+%5Cp%7BNames_List_Subheader%3DVoicing+marks%7D+%5D%0D%0A%26+%5Cp%7BInherited%7D&ucd=on&g=&i=Sc%2CScx%2CEA

[UnicodeSet link]: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%0D%0A++%5Cp%7BScx%3DHira%7D+%5Cp%7BScx%3DKana%7D+%5Cp%7BScx%3DHan%7D+%5Cp%7BScx%3DBopo%7D+%5Cp%7BScx%3DHang%7D%0D%0A++%5B+%5Cu3000-%5Cu303F+%5Cu3190-%5Cu319F+%5Cu31C0-%5Cu31EF+%5Cu3200-%5Cu32FF+%5Cu3300-%5Cu33FF+%5CuFE10-%5CuFE1F+%5CuFE30-%5CuFE4F+%5CuFE50-%5CuFE6F+%5CuFF00-%5CuFFEF+%5D%0D%0A%5D+-+%5B%0D%0A++%5Cp%7BHan%7D+%5Cp%7BKana%7D+%5Cp%7BHira%7D+%5Cp%7BBopo%7D+%5Cp%7BYi%7D%0D%0A++%5B+%5Cp%7BCommon%7D+%26+%5B%5Cp%7BEA%3DH%7D%5Cp%7BEA%3DF%7D%5Cp%7BEA%3DW%7D%5D+-+%5Cp%7BEmoji%7D+-+%5B%E2%82%A9%5D+%5D%0D%0A%5D%0D%0A-+%5Cp%7BUnknown%7D+-+%5Cp%7BHang%7D&g=EA&i=Sc%2CScx%2CEA

isuffix:
I've looked into these character groups some more and I've updated the implementation based on them. Here are my changes for each set and some of my thoughts about their East Asian Width (`EA`) and Script Extension (`Scx`) properties along with the Unicode standard's description for some of them.

### 1. Fullwidth latin letters

I added the fullwidth latin letters. It feels odd that they don't have `Scx=Han` (or something) to indicate they may be used in a CJK context outside of which block they're a part of.


### 2. Circled numbers on black squares

I added the circled numbers on black squares. I have no clue why these would be `EA=Ambiguous` instead of `Wide`. I cannot imagine why a font would display a _square_ containing _two digits_ as a narrow character.

Here is Unicode's description of these characters [(section 22.10.2, paragraph 2)](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G37527):
> The enclosed symbols in the range U+3248..U+324F, which consist of circled numbers ten through eighty on white circles centered on black squares, are encoded for compatibility with the Japanese television standard, ARIB STD B24. In that standard, they are intended to represent symbols for speed limit signs, expressed in kilometers per hour.


### 3. The Six CJ-specific combining marks

I added the six combining marks. These only slipped by the initial definition because they're in `Script=Inherited` instead of `Script=Common`. I [checked], and all other `Inherited` codepoints are `EA=Ambiguous` or `EA=Neutral`.

[checked]: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BInherited%7D%0D%0A-+%5Cp%7BEA%3DN%7D%0D%0A-+%5Cp%7BEA%3DA%7D%0D%0A&g=EA&i=Scx

It may be worth using language information to make determinations on other `Script=Inherited` characters in the future (mostly combining marks and variation selectors), or including them in grapheme cluster analysis, but that will require more research.


### 4. Chinese corner tone marks

I'm ignoring the corner tone marks since they're not for Chinese typesetting, but for transcription of Chinese into English, which is why their `Scx` is `Han|Latin`. They also seem to be largely historical. [Here's a document from 1841](https://archive.org/details/chinesechrestoma00brid_1/page/n19/mode/2up) which uses them.

Here is Unicode's description of these marks [(section 7.8.2, paragraph 2)](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-7/#G17009):
> The characters in the range U+A700..U+A707 are corner tone marks used in the transcription of Chinese. They were invented by Bridgman and Wells Williams in the 1830s. They have little current use, but are seen in a number of old Chinese sources.


### 5. 〿 U+303F IDEOGRAPHIC HALF FILL SPACE

(Note that the character is not failing to render, it is supposed to look like a vertical box with an X through it.)

I'm not adding this character as it doesn't actually seem to be used in Chinese or Japanese writing.

#### Conflicting Properties

This character gave me a lot of trouble because it has two seemingly conflicting property values in Unicode: `Script_Extensions=Han` and `East_Asian_Width=Neutral`. In UAX #11, the [definition](https://www.unicode.org/reports/tr11/tr11-44.html#ED7) of `Neutral` is:
> _Neutral (Not East Asian):_ All other characters. Neutral characters do not occur in legacy East Asian character sets. By extension, they also do not tend to occur in East Asian typography.

It makes no sense for a character to have "no basis in East Asian typography" but also be used in the Han script! 

It's also very confusing that 〿 U+303F is adjacent to 〾 U+303E IDEOGRAPHIC VARIATION INDICATOR, which is a **wholly unrelated character** used in [ideographic description sequences](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-18/#G28626). Trying to look for info about 〿 U+303F down that route is a dead end. But to make it worse, Unicode gives these two a custom subheader in their block as _Special CJK Indicators_ despite having no historical similarity!

Judging by its name alone you would think, "oh, this is used in ideographic contexts, of course it should have `Scx=Han`, so `EA=Neutral` is the issue." But I think that IDEOGRAPHIC in its name is actually a misnomer. Let me explain.

#### Origins

The ideographic half fill space is a very old character in Unicode. It was present even in Unicode 1.0, and while I could not find any records documenting its choice of position in the _CJK Symbols and Punctuation_ block, I did find that its real name used to be "Box X (DBCS Fill Character)" and its real codepoint was 000B when it lived in the IBM code pages for various CJK DOS systems using Double Byte Character Sets.

You can find this from the character's GCGID of `SP500000` [on this page](https://ccsids.net/3-3220-055/gcgids_S.txt#:~:text=SP500000) from [ccsids.net](https://ccsids.net/). And the character is visible in several code pages on Wikipedia such as [page 1040](https://en.wikipedia.org/wiki/Code_page_1040) or [page 1043](https://en.wikipedia.org/wiki/Code_page_1043) or [the first two rows of page 897](https://en.wikipedia.org/wiki/JIS_X_0201#Variants_and_extensions). Although I only got to those places by first finding it in the legacy [external mapping tables (p121)](https://www.unicode.org/versions/Unicode1.0.0/ch06.pdf) for [Unicode 1.0](https://www.unicode.org/versions/Unicode1.0.0/).

When you consider that the DOS code pages were originally designed for terminal-based systems using monospace fonts where characters would render in either one vertical cell or two (based on using either one or two bytes), Unicode's description of the ideographic half fill space starts to make more sense [(section 6.2.13, paragraph 3)](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-6/#G3673):
> U+303F IDEOGRAPHIC HALF FILL SPACE is a visible indicator of a display cell filler used when ideographic characters have been split during display on systems using a double-byte character encoding. It is included in the Unicode Standard for compatibility.

(This description has been present since Unicode 2.0.0 and has not changed)

The character is literally just an X in a box for filling empty cells! It's only "ideographic" because it was only used in code pages for CJK systems, but in that usage it was really a [block element character](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BBlock%3DBlock+Elements%7D) similar to █ U+2588 FULL BLOCK or ▓ U+2593 DARK SHADE (the latter usually sharing an IBM code page with 〿 U+303F at position 0014).

#### Resolution

So of the two conflicting properties, only `EA=Neutral` makes sense! The character does render as narrow, and it doesn't really have a use in East Asian typography in the modern day. It's really `Scx=Han` that's incorrect. And if you look at the data files for `Script_Extensions`, 〿 U+303F is given `Scx=Han` [as part of a range with 〾 U+303E](https://www.unicode.org/Public/UCD/latest/ucd/ScriptExtensions.txt#:~:text=303F), so they seemingly weren't considered separately.

And if you _really_ go looking, you can find reference to it in [revision 3 of UTR #11](https://www.unicode.org/reports/tr11-3/#:~:text=303F)  (June 1999), where Asmus Freytag notes that 〿 U+303F is an exception for an old paragraph in the Unicode standard, as it is a non-full-width character in the set of given blocks, and he updates its `East_Asian_Width` from `Wide` to `Neutral`! But no explanation of its exceptionality is given, and the Background section was removed from later revisions of UTR #11 (which eventually became UAX #11). The closest I could find to a mention of it is in [the meeting minutes for UTC #78 (p10)](https://www.unicode.org/L2/L1998/98419.pdf) (Dec 1998) with some discussion of accepting the—at-the-time draft—UTR as a full technical report. I also found [an email with editorial comments](https://www.unicode.org/mail-arch/unicode-ml/Archives-Old/UML012/0396.html) by John Cowan from that era (June 1998), but no mention of 〿 U+303F. Freytag's [original draft paper for the 'Width' property](https://www.unicode.org/L2/L1998/98054.pdf) may also be of interest.

So I might submit a proposal to have `Scx=Han` removed from U+303F and to note somewhere its similarity to the block element characters.

### Won sign

I also found some more context on the won sign in [UAX #11](https://www.unicode.org/reports/tr11/tr11-44.html#ED3):
> **Note:** Unlike U+00A5 ¥ YEN SIGN, U+20A9 ₩ WON SIGN has an explicit East_Asian_Width property value of East Asian Halfwidth (H). What makes U+00A5 different is that this character was included in a very common—and non–East Asian—character set standard, specifically ISO/IEC 8859-1, and encoded at 0xA5. Almost all legacy Latin fonts supported ISO/IEC 8859-1 in its entirety, using variable-width glyphs. By contrast, most legacy font implementations used an explicit half-width glyph for the won sign, whose source is the standard KS X 1003, and encoded at 0x5C. The assignment of the East Asian Halfwidth (H) property value does not preclude font developers from using a variable-width glyph for U+20A9, and doing so has become a common practice.

This distinction was also discussed many years prior in [the meeting minutes for UTC #78 (p10)](https://www.unicode.org/L2/L1998/98419.pdf), where the answer was: _"Because ISO8859-1 has Yen sign. No other character set has legacy half-width Won sign."_

To be honest, it actually makes me a little mad that the won sign isn't `Ambiguous`, and I may submit a proposal to change it. U+20A9 ₩ is the only character with `EA=Halfwidth` whose name doesn't start with "HALFWIDTH" as part of the _Halfwidth And Fullwidth Forms_ block. I understand that it does have an explicit fullwidth form like the other halfwidth characters, but if its exceptionality has to be called out while other `Ambiguous` characters already need to be specialized by implementations, then what good is it actually doing with its `Halfwidth` designation?

YDX-2147483647:
> 1. Fullwidth latin letters
>
> I added the fullwidth latin letters. It feels odd that they don't have `Scx=Han` (or something) to indicate they may be used in a CJK context outside of which block they're a part of.

I think fullwidth Latin letters should be YesOrAmbiguous.

- Indeed, they are usually used in CJK context. However, even in CJK context, they are still used to write Latin languages. In other words, spaces between fullwidth Latin letters should be kept.

- Some OCR softwares do not distinguish between narrow and fullwidth Latin letters when a Chinese article contains a few English phrases and sentences. The software might output fullwidth characters even if the letters look narrow. (One may still enable the `pwid` OpenType feature to display the fullwidth letters as narrow.) In that case, spaces between them should be kept.

isuffix:
Thank you, I was not aware of that context. I've reverted the fullwidth letters.

laurmaedje (approved):

PR #183, opened 2026-08-26
Add support for styling U+2126 OHM SIGN

MDLC01:
U+2126 OHM SIGN canonically decomposes to the regular capital Omega Greek letter. For this reason, I believe it should behave just like the regular capital Omega with regard to styling.

### DIFF
--- a/src/styling.rs
+++ b/src/styling.rs
@@ -230,6 +230,7 @@ impl MathStyle {
     ) -> MathStyle {
         use MathVariant::*;
         use conversions::*;
+        let c = normalize(c);
         match (variant.unwrap_or(Plain), bold, italic) {
             (SansSerif, false, Some(false)) if is_latin(c) => MathStyle::SansSerif,
             (SansSerif, false, _) if is_latin(c) => MathStyle::SansSerifItalic,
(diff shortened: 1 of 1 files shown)


mkorje (commented):
IMO it'd be cleaner to add a new function like `conversions::normalize`, which converts the ohm sign to captial omega, and then call that at the top of `MathStyle:select` and `to_style`.

MDLC01:
I did that. Also, I was not able to find any other character that we might want to normalize (i.e., a character that canonically decomposes to a character that we can style), but I did not do any form of exhaustive check.

mkorje (approved):

Enivex:
U+212A maps to a regular K

https://www.compart.com/en/unicode/U+212A

There's also all the precomposed letters. For instance, `bold()` does not work on Å (precomposed), but it does work* on Å (using combining character).

*bold is currently broken when applied to combining accents, which I will report if it hasn't been already

Enivex:
Another one: U+1FBE (Greek Prosgegrammeni) decomposes to a regular iota.

(Though I question if anyone would actually use that...)

If Google's AI result isn't lying to me, that should be everything stylable that's not CJK or contains accents.

MDLC01:
I added support for U+212A and U+1FBE.

mkorje (approved):
Re-approving as its stale.

